PR #12, opened 2018-12-05
SLIM-1711 Solve UnrecognizedMessageDataException

renevanr:
SLIM-1711 Debugging showed that buffer.readableBytes() became something like 2147483641. Somehow it overflows or underflows.
Doing the math (writerIndex minus readerIndex) locally within the method yields the correct result.


renevanr:
retest this please

PR #54, opened 2019-03-28
Feature/oc 511 generate correlation

rlemmers:
- Creates a new correlationId for each message sent to core by the protocol adapter.
Generating the correlationId needed the CorrelationIdProvider interface and classes to be moved from org.opensmartgridplatform.domain.core.services to org.opensmartgridplatform.shared.domain.services. Otherwise, the protocol adapter would need a dependency to domain-core.
- Removes messageType from ProtocolResponseMessage, because it's already present in its parent ResponseMessage.
- Removes unused constructor for DeviceMessageMetadata.


osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/infra/jms/ProtocolResponseMessage.java
```diff
@@ -20,7 +20,6 @@
 
     private final String domain;
     private final String domainVersion;
-    private final String messageType;
```
kevinsmeets: The field `messageType` is present in base class `ResponseMessage`, OK.

osgp/protocol-adapter-iec60870/osgp-protocol-adapter-iec60870/src/main/java/org/opensmartgridplatform/adapter/protocol/iec60870/infra/messaging/BaseResponseEventListener.java
```diff
@@ -53,10 +57,19 @@ public void connectionClosed(final IOException e) {
...
         final DeviceMessageMetadata deviceMessageMetadata = new DeviceMessageMetadata(this.messageMetadata);
-        final ProtocolResponseMessage protocolResponseMessage = new ProtocolResponseMessage.Builder()
+
+        final ProtocolResponseMessage.Builder protocolResponseMessageBuilder = new ProtocolResponseMessage.Builder()
                 .domain(this.messageMetadata.getDomain()).domainVersion(this.messageMetadata.getDomainVersion())
                 .deviceMessageMetadata(deviceMessageMetadata).result(ResponseMessageResultType.OK)
-                .retryCount(this.messageMetadata.getRetryCount()).dataObject(response).build();
+                .retryCount(this.messageMetadata.getRetryCount()).dataObject(response);
+
+        LOGGER.info("correlationIdProviderService: {}", this.correlationIdProviderService);
```
kevinsmeets: What is the point in logging this field? Remove it.

rlemmers: I've removed it.

Background: at some point in time, I had an issue with correlationIdProviderService being null. The logging statement was added to check the value. I should have removed it, when I fixed my dependency injection issue.

PR #58, opened 2019-04-03
OC-512 ~ Adds additional condition to prevent multiple notifications

smvdheijden:
[@SkipTests] as it is only a small fix in code that is not being tested yet...


osgp/protocol-adapter-iec60870/osgp-protocol-adapter-iec60870/src/main/java/org/opensmartgridplatform/adapter/protocol/iec60870/infra/messaging/processors/GetHealthStatusResponseEventListener.java
```diff
@@ -44,7 +45,11 @@ public void newASdu(final ASdu receivedAsdu) {
         LOGGER.info("Received the following ASDU for GetHealthStatus: {}", receivedAsdu);
         this.saveReceivedMessage(receivedAsdu);
 
-        if (receivedAsdu.getTypeIdentification() == TypeId.C_IC_NA_1) {
+        // For now return only a response on ASDU with type id: "interrogation
+        // command" and cause of transmission: "activation confirmation".
+        // Processing will change in a later story.
+        if (TypeId.C_IC_NA_1.equals(receivedAsdu.getTypeIdentification())
```
rlemmers: == is shorter and makes it easier to read than equals. We're comparing enum values, so it's okay to use ==.
(applies to both parts of the && clause)

PR #68, opened 2019-05-02
FLEX-4854 ~ Adds OSLP confirm device registration step.

kevinsmeets:


console-test-result-rules
```diff
@@ -1 +1,2 @@
 warning /Failed scenarios/
+warning /Undefined scenarios/
```
kevinsmeets: Makes sure undefined scenarios cause the build result to change to warning.

integration-tests/cucumber-tests-platform/src/test/java/org/opensmartgridplatform/cucumber/platform/PlatformDefaults.java
```diff
@@ -176,6 +176,7 @@
     public static final String LOCALHOST = "127.0.0.1";
     public static final Integer OSLP_SERVER_PORT = 12121;
     public static final Integer RANDOM_DEVICE = 0;
+    public static final Integer RANDOM_PLATFORM = 0;
```
rlemmers: Which code is using this value?

kevinsmeets: No code is using the value (anymore).
I used it to set up the 'confirm device registration' function. Later, I realized it can't work, since the 'device registration' step has created 2 random values for 'random device' and 'random platform'.

PR #101, opened 2019-08-15
SLIM-2049 add missing event codes backend

coendamen:


osgp/platform/osgp-domain-core/src/main/java/org/opensmartgridplatform/domain/core/valueobjects/smartmetering/EventType.java
```diff
@@ -96,16 +96,33 @@
...
+    MANUFACTURER_SPECIFIC_243(243),
+    MANUFACTURER_SPECIFIC_244(244),
+    MANUFACTURER_SPECIFIC_245(245),
+    MANUFACTURER_SPECIFIC_246(246),
+    MANUFACTURER_SPECIFIC_247(247),
+    MANUFACTURER_SPECIFIC_248(248),
+    MANUFACTURER_SPECIFIC_249(249);
 
-    private static final EventType[] VALUES = EventType.values();
-    private static final Map<Integer, EventType> lookup = new HashMap<>();
```
akoster: IMO `lookup` is a better name than `eventCodes`. The Map does not contain eventCodes, but eventTypes and offers a lookup by their value. That said, I think a Map is not very useful here. Iterating over the values to find one is fine for an enum so the static block and the map could be removed and the getValue(int) method can be implemented with an iteration:

```
    for(EventType eventType : EventType.values()) {
        if (eventType.getValue() == intValue) {
            return eventType;
        }
    }
    throw new IllegalArgumentException("No EventType found for intValue " + intValue);
```

Note that the exception above is new; the current implementation simply returns null if the EventType is not found. If feasible the calling code should be refactored to handle such an exception. Another option is to return an Optional<EventType> instead of throwing an exception:

```
    public static Optional<EventType> getValue(final int intValue) {
        for(EventType eventType : EventType.values()) {
            if (eventType.getValue() == intValue) {
                return Optional.of(eventType);
            }
        }
       return Optional.empty();
    }
```

Returning null is a code smell.

osgp/shared/osgp-ws-smartmetering/src/main/resources/schemas/sm-management.xsd
```diff
@@ -506,6 +506,42 @@
...
+      <xsd:enumeration value="MANUFACTURER_SPECIFIC_244"/>
+      <!-- 245 - Indicates a manufacturer specific code -->
+      <xsd:enumeration value="MANUFACTURER_SPECIFIC_245"/>
+      <!-- 246 - Indicates a manufacturer specific code -->
+      <xsd:enumeration value="MANUFACTURER_SPECIFIC_246"/>
+      <!-- 247 - Indicates a manufacturer specific code -->
+      <xsd:enumeration value="MANUFACTURER_SPECIFIC_247"/>
+      <!-- 248 - Indicates a manufacturer specific code -->
+      <xsd:enumeration value="MANUFACTURER_SPECIFIC_248"/>
+      <!-- 249 - Indicates a manufacturer specific code -->
```
akoster: The comments don't add value IMO. I would remove them.

osgp/platform/osgp-domain-core/src/main/java/org/opensmartgridplatform/domain/core/valueobjects/smartmetering/EventType.java
coendamen: you are right, it contains eventTypes and not eventCodes. I mistakenly used codes because that was in the JIRA name.

PR #102, opened 2019-08-20
Feature/slim 2048 mbus channel for smr5

coendamen:


osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/mbus/IdentificationNumberFactory.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private final CommandExecutorDeviceContext deviceContext;
+
+    private IdentificationNumberFactory(final CommandExecutorDeviceContext deviceContext) {
+        this.deviceContext = deviceContext;
+    }
+
+    public static IdentificationNumberFactory create(CommandExecutorDeviceContext deviceContext) {
+        return new IdentificationNumberFactory(deviceContext);
+    }
```
akoster: This is quite a neat pattern, creating a device-specific factory to encapsulate protocol specific logic. However there are some drawbacks:

- the whole DlmsDevice is passed in, but only the protocol (basically DSMR4 vs DMR5) is needed. Too much information is shared
- the logic needs to use conditions throughout to handle the different cases depending on the protocol

So I think we can do better than this. I suggest creating separate classes for the DSMR4 / SMR5 specific logic with an abstract super class which applies the Template Method pattern. See for an example (created by @stefanermens after a similar review comment by me on a PR of his) `org.opensmartgridplatform.adapter.protocol.dlms.domain.commands.dlmsobjectconfig.DlmsObjectConfig`

There needs to be a factory which determines which implementation to use based on the protocol and then the drawbacks mentioned above are avoided.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/mbus/IdentificationNumberFactoryTest.java
```diff
@@ -0,0 +1,116 @@
...
+    @Test
+    public void testFromNullLast8DigitsSmr51() {
+
+        IdentificationNumber identificationNumber = IdentificationNumberFactory.create(
+                createDeviceContext(Protocol.SMR_5_1)).fromLast8Digits(null);
+
+        assertNull(identificationNumber.getIdentificationNumber());
+    }
+
+    private CommandExecutorDeviceContext createDeviceContext(Protocol protocol) {
```
akoster: You should not create a real CommandExecutorDeviceContext in this test, but use a mock instead

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/CommandExecutorDeviceContext.java
```diff
@@ -0,0 +1,35 @@
...
+import org.opensmartgridplatform.adapter.protocol.dlms.domain.entities.DlmsDevice;
+import org.opensmartgridplatform.adapter.protocol.dlms.domain.entities.Protocol;
+
+/**
+ * Contains the current @{@link DlmsDevice} to be used by one single @{@link CommandExecutor} session.
+ *
+ * This class prevents passing device instances across lower level methods when evaluating the Protocol + Versions
+ * (e.g. SMR / 4, 5)
+ */
+public class CommandExecutorDeviceContext {
```
akoster: The class name `CommandExecutorDeviceContext` is not succinct since there is nothing specific to CommandExecutors in this class, it's just a `DeviceContext` or `DeviceHolder`.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/mbus/ScanMbusChannelsCommandExecutor.java
```diff
@@ -106,11 +108,13 @@ public Void fromBundleRequestInput(final ActionRequestDto bundleInput) throws Pr
     public ScanMbusChannelsResponseDto execute(final DlmsConnectionManager conn, final DlmsDevice device,
             final Void mbusAttributesDto) throws OsgpException {
 
+        deviceContext = new CommandExecutorDeviceContext(device);
```
akoster: Same comment here; this is not thread safe

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/CommandExecutorDeviceContext.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public CommandExecutorDeviceContext(DlmsDevice dlmsDevice) {
+        this.dlmsDevice = dlmsDevice;
+    }
+
+    public DlmsDevice getDlmsDevice() {
+        return dlmsDevice;
+    }
+
+    public boolean isSMR5() {
```
akoster: There seems to be no unit test for this class

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/mbus/IdentificationNumber.java
```diff
@@ -26,108 +27,38 @@
...
      * @return a DataObject with the double-long-unsigned value of the
      *         identification number
      */
     public DataObject asDataObject() {
         if (StringUtils.isBlank(this.last8Digits)) {
             return DataObject.newNullData();
         }
-        return DataObject.newUInteger32Data(calculateIdentification(this.last8Digits));
+        return DataObject.newUInteger32Data(getIdentificationNumber());
     }
```
akoster: It looks like there is no unit test for this class.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/mbus/CoupleMBusDeviceCommandExecutor.java
```diff
@@ -87,7 +92,7 @@ public MbusChannelElementsResponseDto execute(final DlmsConnectionManager conn,
          * M-Bus slave device.
          */
         final ChannelElementValuesDto updatedChannelElementValues = this.deviceChannelsHelper
-                .writeUpdatedMbus(conn, requestDto, emptyChannelMatch.getChannel());
+                .writeUpdatedMbus(conn, requestDto, emptyChannelMatch.getChannel(), deviceContext);
```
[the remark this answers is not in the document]
coendamen: the idea was to prevent many methods to have passed the DlmsDevice, avoiding an increase in parameters to many methods. If the agreement was made to use inheritance then I will follow that path.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/mbus/CoupleMBusDeviceCommandExecutor.java
```diff
@@ -42,6 +45,8 @@ public MbusChannelElementsResponseDto execute(final DlmsConnectionManager conn,
 
         LOGGER.debug("retrieving mbus info on e-meter");
 
+        deviceContext = new CommandExecutorDeviceContext(device);
```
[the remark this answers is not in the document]
coendamen: you are right, I did not check the caller logic and assumed each executor has it's own scope, because the Tools project acts based on new Executor().

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/mbus/IdentificationNumberFactory.java
coendamen: indeed, if the Context object will be removed, the scope can be narrowed.

As for the condition, there is only one, so it seems a bit overkill to make separate classes for only the regex.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/CommandExecutorDeviceContext.java
coendamen: the idea was to use this DeviceContext object in all CommandExecutors (to be discussed) when making distinction between SMR4/5

PR #104, opened 2019-08-27
SLIM-2046 Smr5 configuration object

akoster:


osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/ActionMapperService.java
```diff
@@ -204,7 +205,6 @@ private void postConstruct() {
         CLASS_TO_MAPPER_MAP.put(ActivityCalendarData.class, this.configurationMapper);
         CLASS_TO_MAPPER_MAP.put(SetConfigurationObjectRequestData.class, this.configurationMapper);
         CLASS_TO_MAPPER_MAP.put(SetAlarmNotificationsRequestData.class, this.configurationMapper);
-        CLASS_TO_MAPPER_MAP.put(SetConfigurationObjectRequestData.class, this.configurationMapper);
```
akoster: Removed duplicate map key. The rest of the changes in this class are reformatting.

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/ActionMapperService.java
```diff
@@ -248,7 +249,7 @@ public BundleMessagesRequestDto mapAllActions(final BundleMessageRequest bundleM
                 if (mapper != null) {
 
                     if (action instanceof MbusActionRequest) {
-                        action = this.verifyAndFindChannelForMbusRequest((MbusActionRequest) action, smartMeter);
```
akoster: Fixed warning. The method returned the action parameter.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/SetConfigurationObjectCommandExecutor.java
```diff
@@ -47,172 +31,35 @@
...
-    private static final List<ConfigurationFlagTypeDto> FLAGS_TYPES_FORBIDDEN_TO_SET = new ArrayList<>();
-
-    static {
-        FLAGS_TYPES_FORBIDDEN_TO_SET.add(ConfigurationFlagTypeDto.PO_ENABLE);
-        FLAGS_TYPES_FORBIDDEN_TO_SET.add(ConfigurationFlagTypeDto.HLS_3_ON_P_3_ENABLE);
-        FLAGS_TYPES_FORBIDDEN_TO_SET.add(ConfigurationFlagTypeDto.HLS_4_ON_P_3_ENABLE);
-        FLAGS_TYPES_FORBIDDEN_TO_SET.add(ConfigurationFlagTypeDto.HLS_5_ON_P_3_ENABLE);
-        FLAGS_TYPES_FORBIDDEN_TO_SET.add(ConfigurationFlagTypeDto.HLS_3_ON_PO_ENABLE);
-        FLAGS_TYPES_FORBIDDEN_TO_SET.add(ConfigurationFlagTypeDto.HLS_4_ON_PO_ENABLE);
-        FLAGS_TYPES_FORBIDDEN_TO_SET.add(ConfigurationFlagTypeDto.HLS_5_ON_PO_ENABLE);
```
akoster: Moved characteristisc of ConfigurationFlagTypeDto from separate maps into the enum itself

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/dlmsobjectconfig/DlmsObjectConfigService.java
```diff
@@ -53,7 +53,7 @@ public DlmsObjectConfigService(final DlmsHelper dlmsHelper, final List<DlmsObjec
     public Optional<AttributeAddressForProfile> findAttributeAddressForProfile(final DlmsDevice device,
             final DlmsObjectType type, final Integer channel, final DateTime from, final DateTime to,
             final Medium filterMedium) {
-        return this.findDlmsObject(Protocol.withNameAndVersion(device.getProtocol(), device.getProtocolVersion()), type,
+        return this.findDlmsObject(Protocol.forDevice(device), type,
```
akoster: Replaced cludgy convenience method `withNameAndVersion` with the more compact `forDevice`.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/entities/Protocol.java
```diff
@@ -36,20 +37,22 @@ public boolean isSelectValuesInSelectiveAccessSupported() {
...
         }
-
         // If no defined protocol matches, then don't throw exception, but return OTHER_PROTOCOL as default.
         // This is preferred because the adapter is a general DLMS adapter.
         return OTHER_PROTOCOL;
     }
 
-    public static boolean isSMR5(final String protocolName, final String protocolVersion) {
-        Protocol deviceProtocol = Protocol.withNameAndVersion(protocolName, protocolVersion);
-        return deviceProtocol == Protocol.SMR_5_0 || deviceProtocol == Protocol.SMR_5_1;
```
akoster: Replaced the lousy static utility method with an enum instance method.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/dlmsobjectconfig/DlmsObjectConfigServiceTest.java
```diff
@@ -79,28 +79,34 @@
...
+        when(this.config422.findObject(DlmsObjectType.MBUS_MASTER_VALUE, null)).thenReturn(Optional.of(
+                this.registerWithChannel));
+        when(this.config422.findObject(DlmsObjectType.INTERVAL_VALUES, Medium.ELECTRICITY)).thenReturn(Optional.of(
+                this.profileE));
+        when(this.config422.findObject(DlmsObjectType.DAILY_LOAD_PROFILE, null)).thenReturn(Optional.of(
+                this.profileCombined));
+        when(this.config422.findObject(DlmsObjectType.DAILY_LOAD_PROFILE, Medium.GAS)).thenReturn(Optional.of(
+                this.profileCombined));
+        when(this.config422.findObject(DlmsObjectType.DAILY_LOAD_PROFILE, Medium.ELECTRICITY)).thenReturn(Optional.of(
+                this.profileCombined));
```
akoster: Above change looks serious but is only reformatting (introduction of `.this` by the onSave settings)

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/ActionMapperService.java
```diff
@@ -268,14 +269,14 @@ private ActionRequestDto performDefaultMapping(final ActionRequest action, final
 
         if (actionValueObjectDto == null) {
             throw new FunctionalException(FunctionalExceptionType.UNSUPPORTED_DEVICE_ACTION,
-                    ComponentType.DOMAIN_SMART_METERING, new RuntimeException("Object: " + action.getClass().getName()
-                            + " could not be converted to " + clazz.getName()));
+                    ComponentType.DOMAIN_SMART_METERING, new RuntimeException(
+                    "Object: " + action.getClass().getName() + " could not be converted to " + clazz.getName()));
```
coendamen: optional, change to String.format

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/AttributeAddressFactory.java
```diff
@@ -0,0 +1,19 @@
+package org.opensmartgridplatform.adapter.protocol.dlms.domain.commands.configuration.service;
+
+import org.openmuc.jdlms.AttributeAddress;
+import org.openmuc.jdlms.ObisCode;
+
+public class AttributeAddressFactory {
```
coendamen: this factory returns always one and the same object. I assume this class will be extended in the near future? if not, it seems overkill to create a factory for just a single object.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/ProtocolServiceLookup.java
```diff
@@ -0,0 +1,31 @@
+package org.opensmartgridplatform.adapter.protocol.dlms.domain.commands.configuration.service;
+
+import java.util.List;
+
+import org.opensmartgridplatform.adapter.protocol.dlms.domain.entities.Protocol;
+import org.opensmartgridplatform.adapter.protocol.dlms.exceptions.ProtocolAdapterException;
+
+public class ProtocolServiceLookup {
```
coendamen: this name is too generic as is resides in the configuration package. Consider a renamed to ConfigurationServiceLookup.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/ProtocolServiceLookup.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    public GetConfigurationObjectService lookupGetService(final Protocol protocol) throws ProtocolAdapterException {
+        return this.lookupProtocolService(GetConfigurationObjectService.class, protocol);
+    }
+
+    public SetConfigurationObjectService lookupSetService(final Protocol protocol) throws ProtocolAdapterException {
+        return this.lookupProtocolService(SetConfigurationObjectService.class, protocol);
+    }
+
```
coendamen: this is a nice concept, but with Spring you can just grab the bean from the applicationContext by @Component(name=GetConfigurationObjectServiceSmr5).

In that case this class is not needed, but in the caller you can say:

@Autowired
private ApplicationContext сontext;

context.getBean("GetConfigurationObjectService"+protcolName/Version.) and cast to the generic GetConfigurationObjectService

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/SetConfigurationObjectCommandExecutor.java
```diff
@@ -47,172 +31,35 @@
...
-            final ConfigurationFlagTypeDto flagType) {
-        for (final ConfigurationFlagDto configurationFlag : flags) {
-            if (configurationFlag.getConfigurationFlagType().equals(flagType)) {
-                return configurationFlag;
-            }
-        }
-        return null;
+            final ConfigurationObjectDto configurationToSet) throws ProtocolAdapterException {
+        final Protocol protocol = Protocol.forDevice(device);
+        final GetConfigurationObjectService getService = this.protocolServiceLookup.lookupGetService(protocol);
```
coendamen: consider removing these lookups by one method (or class for all command executors) that grabs the bean by name from the Spring ApplicationContext.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/ProtocolService.java
```diff
@@ -0,0 +1,11 @@
+package org.opensmartgridplatform.adapter.protocol.dlms.domain.commands.configuration.service;
```
coendamen: ProtocolService is a marker interface which can be replaced by naming the @Component.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/GetConfigurationObjectCommandExecutorTest.java
```diff
@@ -0,0 +1,50 @@
...
+        device.setProtocol(protocol);
+
+        when(this.protocolServiceLookup.lookupGetService(protocol)).thenReturn(this.getService);
+        when(this.getService.getConfigurationObject(this.conn)).thenReturn(this.configurationObjectDto);
+
+        // CALL
+        final ConfigurationObjectDto result = this.instance.execute(this.conn, device, null);
+
+        // VERIFY
+        assertThat(result).isSameAs(this.configurationObjectDto);
```
coendamen: you check that the result is the mock you set in the getConfigurationObject method.

osgp/shared/osgp-dto/src/main/java/org/opensmartgridplatform/dto/valueobjects/smartmetering/ConfigurationFlagsDto.java
```diff
@@ -17,18 +17,18 @@
...
     @Override
     public String toString() {
-        return this.configurationFlag == null ? "Flags[none]" : String.format("Flags%s", this.configurationFlag);
+        return String.format("Flags %s", this.flags);
     }
 
-    public List<ConfigurationFlagDto> getConfigurationFlag() {
-        return new ArrayList<>(this.configurationFlag);
+    public List<ConfigurationFlagDto> getFlags() {
+        return new ArrayList<>(this.flags);
```
coendamen: consider returning a Collections.unmodifiableList(list);

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/GetConfigurationObjectServiceDsmr4.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(GetConfigurationObjectServiceDsmr4.class);
+
+    private static final int NUMBER_OF_CONFIGURATION_OBJECT_ELEMENTS = 2;
+    private static final int INDEX_OF_GPRS_OPERATION_MODE = 0;
+    private static final int INDEX_OF_CONFIGURATION_FLAGS = 1;
+
+    private final DlmsHelper dlmsHelper;
+
+    public GetConfigurationObjectServiceDsmr4(final DlmsHelper dlmsHelper) {
```
coendamen: I think you need to @Autowired the constructor, otherwise at runtime the DlmsHelper will be null.

osgp/shared/osgp-dto/src/main/java/org/opensmartgridplatform/dto/valueobjects/smartmetering/ConfigurationFlagTypeDto.java
```diff
@@ -9,17 +9,55 @@
...
+    HLS_6_ON_P3_ENABLE(null, 11, false),
+    HLS_7_ON_P3_ENABLE(null, 12, false),
+    HLS_6_ON_P0_ENABLE(null, 13, false),
+    HLS_7_ON_P0_ENABLE(null, 14, false);
+
+    private final Integer bitPositionDsmr4;
+    private final Integer bitPositionSmr5;
+    private final boolean readOnly;
+
+    ConfigurationFlagTypeDto(final Integer bitPositionDsmr4, final Integer bitPositionSmr5, final boolean readOnly) {
```
coendamen: this seems like introducing a carthesian product into an object. the enum set seems to be mutually exclusive, consider making two enumsets. This will simplify the get logic.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectServiceTest.java
```diff
@@ -0,0 +1,101 @@
...
+    private ConfigurationObjectDto configurationToSet;
+    @Mock
+    private ConfigurationObjectDto configurationOnDevice;
+
+    @Before
+    public void setUp() {
+        when(this.conn.getDlmsMessageListener()).thenReturn(this.dlmsMessageListener);
+        when(this.conn.getConnection()).thenReturn(this.dlmsConnection);
+
+        this.instance = new SetConfigurationObjectService(this.dlmsHelper) {
```
coendamen: why create an instance with a dummy impl while this is the class under test?

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/ProtocolServiceLookupTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.opensmartgridplatform.adapter.protocol.dlms.domain.entities.Protocol;
+import org.opensmartgridplatform.adapter.protocol.dlms.exceptions.ProtocolAdapterException;
+
+public class ProtocolServiceLookupTest {
```
coendamen: this test is unnecessary because the Spring Framework has built in lookup of beans at runtime.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/GetConfigurationObjectServiceDsmr4.java
```diff
@@ -0,0 +1,118 @@
...
+        final GprsOperationModeTypeDto gprsMode = this.getGprsOperationMode(elements.get(INDEX_OF_GPRS_OPERATION_MODE));
+        final ConfigurationFlagsDto flags = this.getConfigurationFlags(elements.get(INDEX_OF_CONFIGURATION_FLAGS));
+        return new ConfigurationObjectDto(gprsMode, flags);
+    }
+
+    private GprsOperationModeTypeDto getGprsOperationMode(final DataObject gprsMode) throws ProtocolAdapterException {
+        if (gprsMode == null || !gprsMode.isNumber()) {
+            final String message = String.format(
+                    "Expected ConfigurationObject gprsOperationMode as Number, but got: %s", gprsMode);
+            LOGGER.warn(message);
```
coendamen: a warning is logged although an error is thrown.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/GetConfigurationObjectServiceDsmr4.java
```diff
@@ -0,0 +1,118 @@
...
+        return this.getConfigurationObject(resultData);
+    }
+
+    private ConfigurationObjectDto getConfigurationObject(final DataObject resultData) throws ProtocolAdapterException {
+        LOGGER.info("ConfigurationObject ResultData: {}", this.dlmsHelper.getDebugInfo(resultData));
+        final List<DataObject> elements = resultData.getValue();
+        if (elements == null || elements.size() != NUMBER_OF_CONFIGURATION_OBJECT_ELEMENTS) {
+            final String message = String.format("Expected ConfigurationObject ResultData with %d elements, but got %s",
+                    NUMBER_OF_CONFIGURATION_OBJECT_ELEMENTS, elements == null ? "null" : elements.size());
+            LOGGER.warn(message);
```
coendamen: a warning is logged although an error is thrown.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/GetConfigurationObjectServiceDsmr4.java
```diff
@@ -0,0 +1,118 @@
...
+        return protocol == Protocol.DSMR_4_2_2;
+    }
+
+    @Override
+    ConfigurationObjectDto getConfigurationObject(final GetResult result) throws ProtocolAdapterException {
+        final DataObject resultData = result.getResultData();
+        if (resultData == null || !resultData.isComplex()) {
+            final String message = String.format("Expected ConfigurationObject ResultData as Complex, but got: %s",
+                    resultData);
+            LOGGER.warn(message);
```
coendamen: a warning is logged although an error is thrown.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/GetConfigurationObjectServiceDsmr4.java
```diff
@@ -0,0 +1,118 @@
...
+        }
+        final Number number = gprsMode.getValue();
+        return this.getGprsOperationMode(number.intValue());
+    }
+
+    private GprsOperationModeTypeDto getGprsOperationMode(final int number) {
+        return GprsOperationModeTypeDto.forNumber(number).orElseGet(() -> {
+            LOGGER.warn("Expected GPRS operation mode value to be one of {}, but got: {}",
+                    GprsOperationModeTypeDto.values(), number);
+            return null;
```
coendamen: return null is a code smell.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/GetConfigurationObjectServiceDsmr4.java
```diff
@@ -0,0 +1,118 @@
...
+        });
+    }
+
+    private ConfigurationFlagsDto getConfigurationFlags(final DataObject flags) throws ProtocolAdapterException {
+
+        if (flags == null || !flags.isBitString()) {
+            final String message = String.format("Expected ConfigurationObject flags as BitString, but got: %s", flags);
+            LOGGER.warn(message);
```
coendamen: a warning is logged although an error is thrown.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/GetConfigurationObjectServiceSmr5.java
```diff
@@ -0,0 +1,67 @@
...
+import org.springframework.stereotype.Component;
+
+@Component
+public class GetConfigurationObjectServiceSmr5 extends GetConfigurationObjectService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(GetConfigurationObjectServiceSmr5.class);
+
+    private final DlmsHelper dlmsHelper;
+
+    public GetConfigurationObjectServiceSmr5(final DlmsHelper dlmsHelper) {
```
coendamen: using the dlmsHelper will throw a NPE at runtime because it is not autowired.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectService.java
```diff
@@ -0,0 +1,125 @@
...
+    SetConfigurationObjectService(final DlmsHelper dlmsHelper) {
+        this.dlmsHelper = dlmsHelper;
+    }
+
+    public AccessResultCode setConfigurationObject(final DlmsConnectionManager conn,
+            final ConfigurationObjectDto configurationToSet, final ConfigurationObjectDto configurationOnDevice)
+            throws ProtocolAdapterException {
+
+        final DataObject dataObject = this.buildSetParameterData(configurationToSet, configurationOnDevice);
+        LOGGER.info("ConfigurationObject SetParameter Data : {}", this.dlmsHelper.getDebugInfo(dataObject));
```
coendamen: should this be debug or info ?

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectService.java
```diff
@@ -0,0 +1,125 @@
...
+        final AttributeAddress attributeAddress = AttributeAddressFactory.getConfigurationObjectAddress();
+        final SetParameter setParameter = new SetParameter(attributeAddress, dataObject);
+        conn.getDlmsMessageListener().setDescription(
+                "SetConfigurationObject AttributeAddress: " + JdlmsObjectToStringUtil.describeAttributes(
+                        attributeAddress));
+        return this.getAccessResultCode(conn, setParameter);
+    }
+
+    private AccessResultCode getAccessResultCode(final DlmsConnectionManager conn, final SetParameter setParameter) {
+        LOGGER.info("Set ConfigurationObject using SetParameter {}", setParameter);
```
coendamen: should this be debug or info ?

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectService.java
```diff
@@ -0,0 +1,125 @@
...
+                flagToSet -> !flagToSet.getConfigurationFlagType().isReadOnly()).forEach(flagsToSet::add);
+    }
+
+    // Fill missing flagsToSet with flags from configurationOnDevice
+    private void addDeviceFlags(final ConfigurationObjectDto configurationOnDevice,
+            final List<ConfigurationFlagDto> flagsToSet) {
+        configurationOnDevice.getConfigurationFlags().getFlags().forEach(flagOnDevice -> {
+            if (flagsToSet.stream().noneMatch(
+                    flagToSet -> flagToSet.getConfigurationFlagType() == flagOnDevice.getConfigurationFlagType())) {
+                flagsToSet.add(flagOnDevice);
```
coendamen: consider changing to List.addAll with filter to simplify this statement.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectService.java
```diff
@@ -0,0 +1,125 @@
...
+        final StringBuilder word = new StringBuilder();
+        final char[] pad = new char[NUMBER_OF_FLAG_BITS];
+        Arrays.fill(pad, '0');
+        word.append(pad);
+        return word;
+    }
+
+    private String toWord(final List<ConfigurationFlagDto> flags) throws ProtocolAdapterException {
+        final StringBuilder sb = this.createEmptyWord();
+        for (final ConfigurationFlagDto flag : flags) {
```
coendamen: consider changing to stream.filter(f -> f.isEnabled).map().Collectors.join()

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectServiceDsmr4.java
```diff
@@ -0,0 +1,63 @@
...
+import org.opensmartgridplatform.adapter.protocol.dlms.exceptions.ProtocolAdapterException;
+import org.opensmartgridplatform.dto.valueobjects.smartmetering.ConfigurationFlagTypeDto;
+import org.opensmartgridplatform.dto.valueobjects.smartmetering.ConfigurationObjectDto;
+import org.opensmartgridplatform.dto.valueobjects.smartmetering.GprsOperationModeTypeDto;
+import org.springframework.stereotype.Component;
+
+@Component
+public class SetConfigurationObjectServiceDsmr4 extends SetConfigurationObjectService {
+
+    public SetConfigurationObjectServiceDsmr4(final DlmsHelper dlmsHelper) {
```
coendamen: will throw NPE at runtime when used.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectServiceSmr5.java
```diff
@@ -0,0 +1,35 @@
...
+import org.opensmartgridplatform.adapter.protocol.dlms.domain.entities.Protocol;
+import org.opensmartgridplatform.adapter.protocol.dlms.exceptions.ProtocolAdapterException;
+import org.opensmartgridplatform.dto.valueobjects.smartmetering.ConfigurationFlagTypeDto;
+import org.opensmartgridplatform.dto.valueobjects.smartmetering.ConfigurationObjectDto;
+import org.springframework.stereotype.Component;
+
+@Component
+public class SetConfigurationObjectServiceSmr5 extends SetConfigurationObjectService {
+
+    public SetConfigurationObjectServiceSmr5(final DlmsHelper dlmsHelper) {
```
coendamen: NPE

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/ProtocolServiceLookup.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.List;
+
+import org.opensmartgridplatform.adapter.protocol.dlms.domain.entities.Protocol;
+import org.opensmartgridplatform.adapter.protocol.dlms.exceptions.ProtocolAdapterException;
+
+public class ProtocolServiceLookup {
+
+    private final List<ProtocolService> protocolServices;
+
+    public ProtocolServiceLookup(final List<ProtocolService> protocolServices) {
```
coendamen: I looked at the code, but this class is never instantiated, only in tests

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/AttributeAddressFactory.java
akoster: Not overkill, encapsulation.

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/ActionMapperService.java
```diff
@@ -248,7 +249,7 @@ public BundleMessagesRequestDto mapAllActions(final BundleMessageRequest bundleM
                 if (mapper != null) {
 
                     if (action instanceof MbusActionRequest) {
```
jenkins-ip-10-4-24-184: MAJOR Refactor this code to not nest more than 3 if/for/while/switch/try statements. [rule](http://ci.opensmartgridplatform.org/sonarqube/coding_rules#rule_key=squid%3AS134)

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/SetConfigurationObjectCommandExecutor.java
```diff
@@ -47,172 +31,35 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(SetConfigurationObjectCommandExecutor.class);
```
jenkins-ip-10-4-24-184: MAJOR Remove this unused "LOGGER" private field. [rule](http://ci.opensmartgridplatform.org/sonarqube/coding_rules#rule_key=squid%3AS1068)

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/ActionMapperService.java
akoster: Done.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/ProtocolServiceLookup.java
akoster: The fully qualified name of this class is `org.opensmartgridplatform.adapter.protocol.dlms.domain.commands.configuration.service.ProtocolServiceLookup`. I feel no need to repeat `ProtocolService` in the class name. We have packages for a reason.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/ProtocolServiceLookup.java
akoster: The logic for handling a protocol (see `boolean handles(Protocol)` implementations) is not as straightforward as you suggest. For instance we have SMR 5.0 and 5.1 which should both be handled. Therefore I prefer a single dedicated lookup class than trying to jump through (less understandable) bean lookup hoops with Spring. The latter would also mean this knowledge about which bean names and protocol suffixes are needed gets spread around the code, which is a less maintainable. 

Finally, I like implementation classes themselves to specify which cases they can handle, rather than recording this in some remote location (be it the lookup class, or Spring bean annotations). In this way the lookup class only needs to know that there are a bunch of ProtocolServices which implement Get/Set functionality, nothing about specific protocols or implementation class names.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/SetConfigurationObjectCommandExecutor.java
akoster: I prefer looking up beans by contract than by name, which is what the ProtocolServiceLookup does

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/ProtocolService.java
akoster: ProtolService is not a marker interface. It declares `boolean handles(Protocol)` which allows services to determine which protocols they can handle.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/GetConfigurationObjectCommandExecutorTest.java
akoster: Correct.

osgp/shared/osgp-dto/src/main/java/org/opensmartgridplatform/dto/valueobjects/smartmetering/ConfigurationFlagsDto.java
akoster: I normally don't bother hiding internal collection fields like this because it causes more suffering than happiness IMO (in support, JAXB doesn't do it either). In this case the code was already like this and I decided not to change it to avoid problems. Some calling code might modify the collection after getting it, which would result in bugs when the implementation is changed (either by exposing the list or by returning an unmodifiable copy).

Therefore I decided to leave it as it is.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/GetConfigurationObjectServiceDsmr4.java
akoster: No, Spring Boot does this for us automatically (because DlmsHelper is a Spring bean)

osgp/shared/osgp-dto/src/main/java/org/opensmartgridplatform/dto/valueobjects/smartmetering/ConfigurationFlagTypeDto.java
akoster: The sets of enums are not mutually exclusive, there are three elements they have in common.

We cannot avoid this complexity. We need to record the bit positions for each protocol. There will not be less values if the enum sets are split up, in fact there will be more.

Note that the readOnly value is always false. I left that in the code just in case the requirement to make certain flags read-only comes back.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectServiceTest.java
akoster: Because the class under test is abstract.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/ProtocolServiceLookupTest.java
akoster: This test is only unnecessary if I remove the ProtocolServiceLookup, which is not the case.

Actually this is more evidence that having this lookup is a Good Idea. If the beans were hard wired by name it would be much harder to test the lookup logic.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/GetConfigurationObjectServiceDsmr4.java
akoster: I agree this is bad practice. It was even worse, the messages were not the same...

I could remove the logging but thought it might be handy to quickly diagnose problems on the playground. (I don't know where the ProtocolAdapterException and its message end up).

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/GetConfigurationObjectServiceDsmr4.java
akoster: Fixed.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/GetConfigurationObjectServiceSmr5.java
akoster: This is not true, see above.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectService.java
akoster: Apparently we want to log debugInfo from the dlmsHelper at INFO level.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectService.java
akoster: I'm not aware of a List.addAll with filter method in Java. Could you provide a code snippet to show what you mean?

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectService.java
akoster: That would mean wrapping the checked ProtocolAdapterException in a RuntimeException which seems too much overhead to me.

```
        flags.stream().filter(ConfigurationFlagDto::isEnabled).forEach(flag -> {
            final ConfigurationFlagTypeDto flagType = flag.getConfigurationFlagType();
            final Integer bitPosition = this.getBitPosition(flagType).orElseThrow(() -> new ProtocolAdapterException(
                    String.format("ConfigurationFlagTypeDto %s not known for protocol", flagType)));
            sb.setCharAt(bitPosition, '1');
        });
```

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectServiceDsmr4.java
akoster: Nope.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectServiceSmr5.java
akoster: Nope.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/ProtocolServiceLookup.java
akoster: It is injected by Spring.

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/ActionMapperService.java
akoster: Done

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/SetConfigurationObjectCommandExecutor.java
akoster: Done.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/ProtocolServiceLookup.java
coendamen: but where? ProtocolServiceLookup is not a Spring bean, and the user GetConfigurationObjectCommandExecutor has no autowired constructor nor does it autowire this lookup class.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectServiceDsmr4.java
coendamen: Can you explain? the constructor does not autowire the dlmshelper.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectService.java
coendamen: but why? in the world. a lot of debug code is added as info because most app servers run on info as lowest level.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/GetConfigurationObjectServiceDsmr4.java
coendamen: you don't need to remove the logging, but changing warn to error would be better.

osgp/shared/osgp-dto/src/main/java/org/opensmartgridplatform/dto/valueobjects/smartmetering/ConfigurationFlagTypeDto.java
coendamen: agreed.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/GetConfigurationObjectServiceDsmr4.java
coendamen: ok, I did not know this. with regular spring you need to do this. but Spring Boot is different ?

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectServiceDsmr4.java
coendamen: Starting with Spring 4.3, if a class, which is configured as a Spring bean, has only one constructor, the Autowired annotation can be omitted and Spring will use that constructor and inject all necessary dependencies.

--> I did not know this feature ! so forget previous comments

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/ActionMapperService.java
```diff
@@ -228,54 +228,67 @@ private void postConstruct() {
...
-            }
+        for (final ActionRequest action : bundleMessageRequest.getBundleList()) {
+            actionValueObjectDtoList.add(this.mapActionWithMapper(smartMeter, action));
         }
         return new BundleMessagesRequestDto(actionValueObjectDtoList);
     }
 
+    private ActionDto mapActionWithMapper(final SmartMeter smartMeter, final ActionRequest action) throws FunctionalException {
+        @SuppressWarnings("unchecked")
+        // TODO: fix this
```
jenkins-ip-10-4-24-184: INFO Complete the task associated to this TODO comment. [rule](http://ci.opensmartgridplatform.org/sonarqube/coding_rules#rule_key=squid%3AS1135)

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/SetConfigurationObjectCommandExecutor.java
coendamen: you don't need to return the impl, just return the contract (interface/abstract class)

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/GetConfigurationObjectServiceDsmr4.java
akoster: Indeed.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/GetConfigurationObjectServiceDsmr4.java
akoster: Oh, I thought you were (correctly IMO) pointing out that it's a bad practice to log AND throw exceptions. I'd be happier to remove the logging (after this feature is tested successfully) than to promote it to ERROR level.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectService.java
akoster: Ok, I will lower the level to DEBUG (or remove it completely) after the feature is tested succesfully.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/SetConfigurationObjectService.java
akoster: @coendamen could you give an example of what you mean?

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/service/ProtocolServiceLookup.java
akoster: You are right, I forgot the @Component annotation on the ProtocolServiceLookup. Has been discovered and fixed by me in the meantime.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/configuration/SetConfigurationObjectCommandExecutor.java
coendamen: autowire both implementations by the contract, based on the protocol use either the DSMR_4 or SMR_5 impl. The lookup, and the protocolservice classes can be removed because they are just a wrapper around the Spring context.

PR #121, opened 2019-10-18
FLEX-5068: Enables parsing of date strings for the event date time.

robindenadel:


integration-tests/cucumber-tests-platform/src/test/java/org/opensmartgridplatform/cucumber/platform/glue/steps/database/core/EventSteps.java
```diff
@@ -52,14 +52,22 @@ public void anEvent(final Map<String, String> data) {
...
     }
 
+    private DateTime getDateTime(final String dateTime) {
+        try {
+            return getDateTime2(dateTime, DateTime.now());
+        } catch (final IllegalArgumentException e) {
+            return DateTime.parse(dateTime);
+        }
+    }
+
```
rlemmers: Extend the DateTimeHelper.getDateTime method with the "DateTime.parse(..)" part. It makes the utility method more generic and prevents adding date/time manipulations in the EventSteps class.

Note: the current version of DateTimeHelper.getDateTime(..) accepts Strings like "today at midnight + 5 hours". It can't handle absolute date/time values yet.

robindenadel: It was clear to me that the current version accepts strings like that. Not very useful in the scenarios I created.

robindenadel: Moved the parsing to the DateTimeHelper.

PR #149, opened 2020-01-17
OC-535 ~ Updates JAXB Maven plugin in jasper wireless

smvdheijden:


osgp/protocol-adapter-dlms/osgp-jasper-interface/pom.xml
```diff
@@ -40,21 +41,18 @@
       <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>jaxb2-maven-plugin</artifactId>
-        <version>1.5</version>
+        <version>2.4</version>
```
robindenadel: retrieve the version from the super pom

PR #153, opened 2020-01-27
FLEX-5239 ~ Upgrade to Cucumber 5

bvdzwet:
Refactors packages of Cucumber classes to the newer API. The core part
is considered internal and to be used through the java and junit
projects.

Add strict equals true to the AcceptanceTests classes as the non-strict
mode is deprecated, and will be removed in a future version. This
prevents warnings in the output logs.

Updates settings handling in the steps, to get rid of issues because of
empty cells defaulting to null instead of the empty string as before.

Updates step method annotations where the step definition text with the
step methods is not wrapped between "^" and "$" and the step has
parameters. These are now considered Cucumber expressions instead of
regular expressions. For Cucumber expressions a different syntax is
needed to match the arguments.

Updates the Configurer to no longer implement TypeRegistryConfigurer.
The configurers registers some default step data transformers without
implementing an interface of which only one implementation is allowed on
the glue path. This should offer flexibility in adding custom
transformer functionality with other step classes as needed, without
having to configure this in a single class.

Adds the junit-vintage-engine to the POMs of projects with Cucumber
tests. The current setup of Cucumber depends on JUnit 4.

Uncommented and at-skipped a feature in
TariffSwitchingResponseDataCleanupJob.feature in
cucumber-tests-platform-publiclighting. A feature file without any
contents leads to errors when running the Cucumber tests.

### DIFF
--- a/integration-tests/cucumber-tests-platform-common/src/test/java/org/opensmartgridplatform/cucumber/platform/common/glue/steps/ws/basicosgpfunctions/AuthorizeDeviceFunctionsSteps.java
+++ b/integration-tests/cucumber-tests-platform-common/src/test/java/org/opensmartgridplatform/cucumber/platform/common/glue/steps/ws/basicosgpfunctions/AuthorizeDeviceFunctionsSteps.java
@@ -52,8 +52,8 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.ws.soap.client.SoapFaultClientException;
 
-import cucumber.api.java.en.Then;
-import cucumber.api.java.en.When;
+import io.cucumber.java.en.Then;
+import io.cucumber.java.en.When;
 
 /**
  * Class with all the AuthorizeDeviceFunctions steps
@@ -152,7 +152,7 @@ public class AuthorizeDeviceFunctionsSteps {
         }
     }
 
-    @Then("the device function response is \"([^\"]*)\"")
+    @Then("the device function response is \"{}\"")
     public void theDeviceFunctionResponseIsSuccessful(final Boolean allowed) {
         if (allowed) {
             final Object response = ScenarioContext.current().get(PlatformCommonKeys.RESPONSE);
@@ -205,8 +205,9 @@ public class AuthorizeDeviceFunctionsSteps {
 
         updateDeviceAuthorisationsRequest.getDeviceAuthorisations().add(deviceAuthorisation);
 
-        ScenarioContext.current().put(PlatformCommonKeys.RESPONSE,
-                this.adminDeviceManagementClient.updateDeviceAuthorisations(updateDeviceAuthorisationsRequest));
+        ScenarioContext.current()
+                .put(PlatformCommonKeys.RESPONSE,
+                        this.adminDeviceManagementClient.updateDeviceAuthorisations(updateDeviceAuthorisationsRequest));
     }
 
     private void startSelfTest(final Map<String, String> requestParameters)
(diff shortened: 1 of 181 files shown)


CLAAdmin:
[CLA assistant check](http://cla.opensmartgridplatform.org:80/OSGP/open-smart-grid-platform?pullRequest=153) <br/>You should sign our Contributor License Agreement in order to get your pull request merged.<br/>

integration-tests/cucumber-tests-platform-common/src/test/java/org/opensmartgridplatform/cucumber/platform/common/glue/steps/ws/basicosgpfunctions/AuthorizeDeviceFunctionsSteps.java
```diff
@@ -152,7 +152,7 @@ public void receivingADeviceFunctionRequest(final Map<String, String> requestPar
         }
     }
 
-    @Then("the device function response is \"([^\"]*)\"")
+    @Then("the device function response is \"{}\"")
```
kevinsmeets: Any matcher for Boolean argument looks like the best alternative, OK.

integration-tests/cucumber-tests-platform-common/src/test/java/org/opensmartgridplatform/cucumber/platform/common/glue/steps/ws/core/deviceinstallation/GetStatusSteps.java
```diff
@@ -71,7 +72,7 @@ public void theDeviceInstallationGetStatusAsyncResponseContains(final Map<String
                         PlatformDefaults.DEFAULT_ORGANIZATION_IDENTIFICATION));
     }
 
-    @Then("the platform buffers a device installation get status response message for device \"([^\"]*)\"")
+    @Then("the platform buffers a device installation get status response message for device {string}")
```
kevinsmeets: The original regex matches a string enclosed in double-quotes. The new Cucumber expression does not take into account the double-quotes. The feature file contains:

And the platform buffers a device installation get status response message for device "TEST1024000000001"

Should the Cucumber expression be changed?

integration-tests/cucumber-tests-platform/src/test/java/org/opensmartgridplatform/cucumber/platform/glue/steps/database/core/ScheduledTaskSteps.java
```diff
@@ -44,7 +44,7 @@
     @Autowired
     private ScheduledTaskRepository scheduledTaskRepository;
 
-    @Given("a scheduled \"([^\"]*)\" task")
+    @Given("a scheduled {string} task")
```
kevinsmeets: The original regex matches a string enclosed in double-quotes. The new Cucumber expression does not take into account the double-quotes.

I can't find any feature files using this step.

integration-tests/cucumber-tests-platform-common/src/test/java/org/opensmartgridplatform/cucumber/platform/common/glue/steps/ws/core/deviceinstallation/StopDeviceSteps.java
```diff
@@ -76,7 +76,7 @@ public void theStopDeviceResponseContainsSoapFault(final Map<String, String> exp
         GenericResponseSteps.verifySoapFault(expectedResult);
     }
 
-    @Then("the platform buffers a stop device response message for device \"([^\"]*)\"")
+    @Then("the platform buffers a stop device response message for device {string}")
```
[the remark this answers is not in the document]
bvdzwet: The replaced text was a regex match for the device identification.
Since the string did not start with ^/end with $ the test is treated as a Cucumber Expression.
I chose to replace the regex with the Cucumber expression alternative {string} since it looks simpler and does the same thing.

integration-tests/cucumber-tests-platform-common/src/test/java/org/opensmartgridplatform/cucumber/platform/common/glue/steps/ws/core/deviceinstallation/GetStatusSteps.java
bvdzwet: The Cucumber expression with {string} does take into account the double quotes (or alternatively single quotes).

integration-tests/cucumber-tests-platform/src/test/java/org/opensmartgridplatform/cucumber/platform/glue/steps/database/core/ScheduledTaskSteps.java
bvdzwet: Most changes were based on failing tests, but near the end I did some regex based searching and found one (I think) more potential regex issue, and replaced it.
I have intentionally been a bit careful about extra changes beside the upgrade, but if you think this had better been removed, I can take it out.

PR #173, opened 2020-02-14
Bug/flex 5258 fix optional get

smvdheijden:


osgp/platform/osgp-adapter-domain-core/src/main/java/org/opensmartgridplatform/adapter/domain/core/application/services/FirmwareManagementService.java
```diff
@@ -147,12 +148,18 @@ private static String getFirmwareFilename(final String firmwareUrl) {
     }
 
     private static FirmwareModuleType getFirmwareModuleType(final Map<FirmwareModule, String> firmwareModuleVersions) {
-        final FirmwareModule firmwareModule = firmwareModuleVersions.keySet().stream().findFirst().get();
-        return FirmwareModuleType.valueOf(firmwareModule.getDescription().toUpperCase());
+        return firmwareModuleVersions.keySet()
+                .stream()
+                .findFirst()
+                .map(fwModuleVersion -> FirmwareModuleType.valueOf(fwModuleVersion.getDescription().toUpperCase()))
+                .orElseThrow(() -> new NoSuchElementException("No entries present in firmwareModuleVersions map."));
```
bvdzwet: No issues with this change on itself.
Trying to understand if the chosen exception is the appropriate way to go, I looked at where this code is called from (createSsldPendingFirmwareUpdateRecord).
Before this method (and getFirmwareVersion from the other fix) is called the map is asserted to have exactly one element. Because of that assertion the exception here will never be triggered, since the stream on the maps entry set (or values) always has a single element.
The code could be simplified by obtaining the first `Entry<FirmwareModule, String>` (which is there because of the earlier assertion) and using its key to determine `firmwareModuleType` and its value as `firmwareVersion`.

That said, the code does what it needs to do. As far as I'm concerned you can check how strongly you (dis)agree about further changes being an improvement, and decide whether to leave it, or update it some more.

PR #279, opened 2020-04-14
FLEX-4693: Removes MD5 functionality.

robindenadel:


osgp/platform/osgp-domain-core/src/main/java/org/opensmartgridplatform/domain/core/entities/FirmwareFile.java
```diff
@@ -95,6 +95,11 @@ public FirmwareFile(final String identification, final String filename, final St
         this.hash = hash;
     }
 
+    public FirmwareFile(final String identification, final String filename, final String description,
```
rlemmers, line 98: The class already had 5 constructors, which were doing more or less the same. Instead of adding two more constructors, use a builder.

PR #342, opened 2020-06-04
FLEX-5303 Handle IEC60870 light measurement events

bvdzwet:
Implements IEC60870 event acquisition scenario for light measurement
devices spontaneously sending an ASDU with data concerning the light
sensor status.
The data with the event is translated to an event notification for the
platform indicating whether the light sensor of the light measurement
device reports dark or light.

Update the default device type from DA_DEVICE (old name of a renamed
enum value) to DISTRIBUTION_AUTOMATION_DEVICE.


osgp/protocol-adapter-iec60870/osgp-protocol-adapter-iec60870/src/main/java/org/opensmartgridplatform/adapter/protocol/iec60870/infra/messaging/OsgpRequestMessageSender.java
```diff
@@ -0,0 +1,54 @@
...
+            public Message createMessage(final Session session) throws JMSException {
+                final ObjectMessage objectMessage = session.createObjectMessage(requestMessage);
+                objectMessage.setJMSType(messageType);
+                objectMessage.setStringProperty(Constants.ORGANISATION_IDENTIFICATION,
+                        requestMessage.getOrganisationIdentification());
+                objectMessage.setStringProperty(Constants.DEVICE_IDENTIFICATION,
+                        requestMessage.getDeviceIdentification());
+
+                return objectMessage;
+            }
```
robindenadel, line 49: we could do without the anonymous inner class:
this.jmsTemplate.send(session -> {
    final ObjectMessage objectMessage = session.createObjectMessage(requestMessage);
    objectMessage.setJMSType(messageType);
    objectMessage.setStringProperty(Constants.ORGANISATION_IDENTIFICATION,
            requestMessage.getOrganisationIdentification());
    objectMessage.setStringProperty(Constants.DEVICE_IDENTIFICATION, requestMessage.getDeviceIdentification());

    return objectMessage;
});

bvdzwet: I replaced the anonymous class by a lambda expression.
It may be better if we figure out if one form or the other is a preferred way of handling things.
Without a declared preference this might just lead to changing "in the other direction" in some other pull request.

PR #344, opened 2020-06-08
FLEX-5395 Add configuration of IP addresses that may be duplicated

bvdzwet:
Upon device registration, the IP address of the registered device is
removed in the database for other devices where it is configured (with
the exception of the loopback address - localhost / 127.0.0.1).

These changes allow to either configure that such clean up should never
be performed (allowing multiple devices to share any network address),
or to configure a comma separated list of IP addresses or IP address
ranges that may be shared (for instance
"192.168.0.1,192.168.1.12-192.168.2.103" to allow sharing of 192.168.0.1
or any IP address from 192.168.1.12 to 192.168.2.103 inclusive).

### DIFF
--- a/osgp/platform/osgp-core/src/test/java/org/opensmartgridplatform/core/application/services/DeviceNetworkAddressCleanupServiceTest.java
+++ b/osgp/platform/osgp-core/src/test/java/org/opensmartgridplatform/core/application/services/DeviceNetworkAddressCleanupServiceTest.java
@@ -1,0 +1,168 @@
+/**
+ * Copyright 2020 Smart Society Services B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.opensmartgridplatform.core.application.services;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import java.net.InetAddress;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.opensmartgridplatform.domain.core.entities.Device;
+import org.opensmartgridplatform.domain.core.repositories.DeviceRepository;
+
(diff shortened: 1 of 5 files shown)


osgp/platform/osgp-core/src/test/java/org/opensmartgridplatform/core/application/services/DeviceNetworkAddressCleanupServiceTest.java
```diff
@@ -0,0 +1,168 @@
...
+
+    private DeviceRepository deviceRepository;
+    private DeviceNetworkAddressCleanupService deviceNetworkAddressCleanupService;
+
+    @BeforeEach
+    public void setUp() {
+        this.deviceRepository = Mockito.mock(DeviceRepository.class);
+        final boolean allowMultipleDevicesPerNetworkAddress = false;
+        final List<String> ipRangesAllowingMultipleDevicesPerAddress = Collections.emptyList();
+        this.deviceNetworkAddressCleanupService = new DeviceNetworkAddressCleanupService(this.deviceRepository,
```
robindenadel, line 44: A bit strange to put this in setUp, while it is overridden in a lot of tests

osgp/platform/osgp-core/src/test/java/org/opensmartgridplatform/core/application/services/DeviceNetworkAddressCleanupServiceTest.java
```diff
@@ -0,0 +1,168 @@
...
+    @Test
+    public void duplicateAddressesAreAllowedForLoopbackAddresses() throws Exception {
+        assertThat(this.deviceNetworkAddressCleanupService.allowDuplicateEntries(InetAddress.getLoopbackAddress()))
+                .isTrue();
+    }
+
+    @Test
+    public void noDevicesAreUpdatedWhenTheNetworkAddressIsNotUsed() throws Exception {
+        final String host = "192.168.0.13";
+        final InetAddress inetAddress = InetAddress.getByName(host);
```
robindenadel, line 58: You could consider making constants of these variables, since they are used in almost every test

bvdzwet: I changed the way the setup is handled, so this default before-each that is hardly ever used is no longer applied.

osgp/platform/osgp-core/src/test/java/org/opensmartgridplatform/core/application/services/DeviceNetworkAddressCleanupServiceTest.java
bvdzwet: I changed the way the setup is handled, so this default before-each that is hardly ever used is no longer applied.

osgp/platform/osgp-core/src/test/java/org/opensmartgridplatform/core/application/services/DeviceNetworkAddressCleanupServiceTest.java
bvdzwet: I changed the way the host and InetAddress are configured in the tests, making some use of constants. In general I don't think we should be too eager to extract as much as possible from test code, as this can make the separate tests harder to understand looking at them in isolation, without tracing other parts of the code. I hope you are OK with the balance I tried to find here.

PR #351, opened 2020-06-11
SLIM-2491 invocation counter

coendamen:


osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/factories/Hls5Connector.java
```diff
@@ -141,9 +141,12 @@ private void configureIvData(final TcpConnectionBuilder tcpConnectionBuilder, fi
             manufacturerId = device.getManufacturerId();
         }
         tcpConnectionBuilder.setSystemTitle(manufacturerId, device.getDeviceId());
-        tcpConnectionBuilder.setFrameCounter(device.getInvocationCounter());
-        LOGGER.debug("Framecounter for device {} set to {}", device.getDeviceIdentification(),
-                device.getInvocationCounter());
+
+        UnsignedInteger frameCounter = UnsignedInteger.valueOf(device.getInvocationCounter());
```
coendamen, line 145: because the API can only handle an Integer. Gerbert/Paul/myself decided to keep it like this, all will be changed to a Long, except for this. A bug report was sent to the jDLMS library maintainers to update to a Long.
I explicitly stated we are using an UnsignedInteger here according to the spec.

PR #364, opened 2020-07-06
Added equals and hashCode impl to serializable class

jboon:


osgp/platform/osgp-domain-core/pom.xml
```diff
@@ -195,6 +195,13 @@
...
     </dependency>
 
+    <!-- Utils -->
+    <dependency>
+      <groupId>org.projectlombok</groupId>
+      <artifactId>lombok</artifactId>
+      <scope>provided</scope>
+    </dependency>
+
     <!-- Testing dependencies -->
```
coendamen, line 205: +1

PR #401, opened 2020-08-17
SLIM-2631 fix connection slot leak

coendamen:


osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/infra/messaging/DlmsConnectionMessageProcessor.java
```diff
@@ -58,13 +57,18 @@
...
             final MessageMetadata messageMetadata) throws OsgpException {
-        
+
         throttlingService.openConnection();
 
         final DlmsMessageListener dlmsMessageListener = this
                 .createMessageListenerForDeviceConnection(device, messageMetadata);
 
-        return this.dlmsConnectionHelper.createConnectionForDevice(device, dlmsMessageListener);
+        try {
```
coendamen, line 66: this is the fix, if the connection fails somehow, release the slot. There is a postProcessing method also that releases the connection, but it checks it the connection was null and then returns, but of course in case of connection failure the connection is null, so the slot was never released.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/infra/messaging/DlmsConnectionMessageProcessor.java
```diff
@@ -58,13 +57,18 @@
...
+            return this.dlmsConnectionHelper.createConnectionForDevice(device, dlmsMessageListener);
+        } catch (Exception e) {
+            throttlingService.closeConnection();
```
jandik, line 69: Use finally instead?

PR #500, opened 2020-11-10
FLEX-5588: Don't send bay identification in case of a meta measurement.

robindenadel:


osgp/platform/osgp-adapter-kafka-distributionautomation/src/test/java/org/opensmartgridplatform/adapter/kafka/da/application/mapping/ScadaMeasurementPublishedEventConverterTest.java
```diff
@@ -105,4 +110,34 @@ void testNullString() {
         assertThat(event).isNull();
     }
 
+    @Test
+    void testEsm() {
```
bvdzwet, line 114: Were this test and the helper method below it added accidentally with the other changes?
If not, it might be good to make it a little clearer and actually assert something instead of printing the length to system out.

robindenadel:
retest this please

PR #526, opened 2020-11-27
FLEX-5474 Add reboot behavior to the OSLP web device simulator

bvdzwet:
When the OSLP web device simulator simulates a reboot a configurable
delay in seconds is applied before the device registration messages are
sent. During this delay the Netty channel for new requests will be
disconnected by the OlspChannelHandler.

The delay has a default value from configuration (reboot.delay.seconds)
and can be changed from the devices page in the web-device-simulator.

Fixes a number of warnings in the DeviceManagementController.
The general RequestMapping with a method is replaces by GetMapping or
PostMapping.
Some checks are done in the API methods that return a response for a
bad request in case there is something wrong with the provided data.
Some of the methods that were altering state have been renamed from get
to set.
The random for generating bytes is placed in a field instead of creating
a new one on each call in the method where it is used.
An incorrect logger statement with an exception is fixed.

The labels in the device list page have been placed in actual label
elements for the controls they belong with (making clicking the check
box inputs simpler as one side effect).

A call to fireChannelRead is removed from the OslpChannelHandler as it
is the last handler in the pipeline. Passing the message on beyond the
last handler triggers some debug level warning messages from Netty.

### DIFF
--- a/osgp/protocol-adapter-oslp/web-device-simulator/src/main/java/org/opensmartgridplatform/webdevicesimulator/service/OslpChannelHandler.java
+++ b/osgp/protocol-adapter-oslp/web-device-simulator/src/main/java/org/opensmartgridplatform/webdevicesimulator/service/OslpChannelHandler.java
@@ -11,15 +11,20 @@ import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.security.PrivateKey;
-import java.text.ParseException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Random;
+import java.util.SortedSet;
 import java.util.Timer;
 import java.util.TimerTask;
+import java.util.TreeSet;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -80,6 +85,9 @@ import io.netty.channel.SimpleChannelInboundHandler;
 public class OslpChannelHandler extends SimpleChannelInboundHandler<OslpEnvelope> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(OslpChannelHandler.class);
+
+    private static final SortedSet<String> REBOOTING_DEVICE_IDS = Collections.synchronizedSortedSet(new TreeSet<>());
+
     private final Lock lock = new ReentrantLock();
     private final ConcurrentMap<String, Callback> callbacks = new ConcurrentHashMap<>();
     private final List<OutOfSequenceEvent> outOfSequenceList = new ArrayList<>();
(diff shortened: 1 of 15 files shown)


osgp/protocol-adapter-oslp/web-device-simulator/src/main/java/org/opensmartgridplatform/webdevicesimulator/service/OslpChannelHandler.java
```diff
@@ -871,13 +919,25 @@ private void handleUpdateFirmwareRequest(final Device device, final UpdateFirmwa
...
+                event = Oslp.Event.FIRMWARE_EVENTS_ACTIVATING;
+                description = "A new firmware is activated";
+            } else {
+                LOGGER.warn("Device {} has issues registering after reboot handling update firmware request",
+                        device.getDeviceIdentification());
+                event = Oslp.Event.FIRMWARE_EVENTS_DOWNLOAD_FAILED;
+                description = "Failure rebooting and registrating device during firmware activation";
+            }
+            this.sendEventWithCustomDelay(device, event, description,
+                    Math.max(1, 7 - this.deviceManagementService.getRebootDelay()));
```
joostknapen, line 939: I see what you are doing here, but on the other hand, this would  be an oppurtunity to get rid of the hard coded 7 second delay for the FIRMWARE_EVENTS_ACTIVATING event. I agree with you, that it's out of scope for this story, but maybe getting the 7 from a property file is not much work. I leave it up to you what to do with it. For me it's fine to leave it as is right now.

osgp/protocol-adapter-oslp/web-device-simulator/src/main/java/org/opensmartgridplatform/webdevicesimulator/web/controller/DeviceManagementController.java
```diff
@@ -177,77 +184,114 @@ private void fetchData(final Model model, final String deviceIdentification, fin
...
         return this.deviceManagementService.getLightSwitching();
     }
 
-    @RequestMapping(value = EVENT_NOTIFICATION_CHECK_URL, method = RequestMethod.POST)
-    public ResponseEntity<String> getEventNotificationValue(@RequestBody final AutonomousRequest request) {
+    @PostMapping(value = EVENT_NOTIFICATION_CHECK_URL)
+    public ResponseEntity<String> setEventNotificationValue(@RequestBody final AutonomousRequest request) {
+        if (request == null || request.getAutonomousStatus() == null) {
+            return this.badRequestResponse(request);
+        }
```
joostknapen, line 264: I see this check 5 times. I would refactor it to a method call.

osgp/protocol-adapter-oslp/web-device-simulator/src/main/webapp/WEB-INF/views/devices/list.jsp
```diff
@@ -73,16 +73,18 @@ body {
...
+                        <input id="devReboot" name="devReboot" type="checkbox" style="display: inline;" />
+                        <label for="tariffSwitching" style="display: inline; margin-left: .5em;"><spring:message code="device.edit.page.device.tariff.switching" /></label>
+                        <input id="tariffSwitching" name="tariffSwitching" type="checkbox" style="display: inline;" />
+                        <label for="lightSwitching" style="display: inline; margin-left: .5em;"><spring:message code="device.edit.page.device.light.switching" /></label>
+                        <input id="lightSwitching" name="lightSwitching" type="checkbox" style="display: inline;" />
+                        <label for="eventListener" style="display: inline; margin-left: .5em;"><spring:message code="device.edit.page.device.event.notification" /></label>
+                        <input id="eventListener" name="eventListener" type="checkbox" style="display: inline;" />
+                        <label for="rebootDelay" style="display: inline; margin-left: .5em;"><spring:message code="device.edit.page.device.reboot.delay" /></label>
```
joostknapen, line 86: It's generally better to put styles in a css class.

osgp/protocol-adapter-oslp/web-device-simulator/src/main/java/org/opensmartgridplatform/webdevicesimulator/web/controller/DeviceManagementController.java
joostknapen: Lease as is

osgp/protocol-adapter-oslp/web-device-simulator/src/main/webapp/WEB-INF/views/devices/list.jsp
joostknapen: Leave as is, since the current css is even worse.

osgp/protocol-adapter-oslp/web-device-simulator/src/main/java/org/opensmartgridplatform/webdevicesimulator/service/OslpChannelHandler.java
joostknapen: In case the 7 sec should be configurable, it should be another user story.

joostknapen (approved):

PR #535, opened 2020-12-09
Feature/slim 2569 actual meter readings

harrymiddelburgetall:
This PR adds additional existing requests and corresponding responses to the EI - External Interface
- GetActualMeterReadsRequest
- GetActualMeterReadsGasRequest
- ActualMeterReadsReponse
- ActualMeterReadsGasResponse


osgp/shared/osgp-ws-smartmetering/src/main/resources/schemas/sm-external-interface.xsd
```diff
@@ -29,6 +29,8 @@
           <xsd:element ref="bun:GetPowerQualityProfileRequest"/>
           <xsd:element ref="adm:SetMaintenanceStateRequest"/>
           <xsd:element ref="bun:GetOutagesRequest"/>
+          <xsd:element ref="bun:GetActualMeterReadsRequest"/>
+          <xsd:element ref="bun:GetActualMeterReadsGasRequest"/>
         </xsd:choice>
       </xsd:sequence>
     </xsd:complexType>
```
bvdzwet, line 36: Regarding the changes I only have a question, and that is if it is intentional to only add the requests while adding nothing to the responses?

Regarding to this file itself, I have another remark. It does not belong in the GXF open-smart-grid-platform repository, as GXF has nothing to do with the external interface.

harrymiddelburgetall: The missing specific response types are handled by their super type Response and therefor not needed to be specified as part in the ExternalInterfaceResultMessage type.  Considering from the symetrics point of view these specific response types will be added.
A story has been put on the Smart Meters backlog to transfer the EI definition to the Smart Meters project (SLIM-2882)

bvdzwet (approved):

PR #550, opened 2021-01-11
FLEX-5513: changed iec61850.ssld.port.server to 60102 instead of 102

joostknapen:


osgp/protocol-adapter-iec61850/osgp-protocol-adapter-iec61850/src/main/resources/osgp-adapter-protocol-iec61850.properties
```diff
@@ -70,7 +70,7 @@ iec61850.timeout.connect=10000
 iec61850.port.client=102
 iec61850.port.clientlocal=12345
 iec61850.rtu.port.server=60102
-iec61850.ssld.port.server=102
+iec61850.ssld.port.server=60102
```
smvdheijden, line 73: Couldn't this lead to issues on environments with real devices using port 102?
I suspect this change is probably not needed anyway, as the simulator is listening on both ports 102 and 60102?

smvdheijden (changes requested):

smvdheijden:
Closing this PR as it should not be needed to change the port server.

PR #558, opened 2021-01-26
SLIM-2876 ~ Makes notifications more robust

rlemmers:


osgp/platform/osgp-adapter-ws-smartmetering/src/main/java/org/opensmartgridplatform/adapter/ws/smartmetering/infra/jms/messageprocessor/DomainResponseMessageProcessor.java
```diff
@@ -135,7 +134,10 @@ public void processMessage(final ObjectMessage message) {
                             correlationUid, String.valueOf(notificationType)));
 
         } catch (final Exception e) {
-            this.handleError(e, correlationUid, organisationIdentification, deviceIdentification, notificationType);
+            // Logging is enough, sending the notification will be done
+            // automatically by the resend notification job
+            LOGGER.warn("Error: {} occurred for notification type: {} with correlationUid: {}", e.getMessage(),
+                    notificationType, correlationUid, e);
```
bvdzwet, line 140: I wonder if the phrasing of the warning message is a bit too severe for what is going on.
Starting with "Error" might trigger someone seeing this in the logs to look into what should be corrected,
As the issue here will fix itself as suggested by the comment in the code, a log line starting with something like "Unable to deliver notification for ..." could be nicer.
As you have looked more into the application logs than I have, I'll leave it up to your judgement whether you think this could be improved or whether it is fine as it is.

osgp/platform/osgp-adapter-ws-smartmetering/src/test/java/org/opensmartgridplatform/adapter/ws/smartmetering/infra/jms/messageprocessor/DomainResponseMessageProcessorTest.java
```diff
@@ -0,0 +1,87 @@
...
+        this.messageProcessor.processMessage(this.message);
+
+        verify(this.responseDataService).enqueue(any());
+        verify(this.notificationService).sendNotification(any(), any());
+    }
+
+    @Test
+    void processMessageErrorInHandleMessage() throws JMSException {
+
+        org.mockito.Mockito.doThrow(new NullPointerException("Some runtime exception"))
```
bvdzwet, line 66: The fully qualified  `org.mockito.Mockito.doThrow` looks a bit out of tone with the other Mockito methods with static imports. When using methods from a single place (`org.mockito.Mockito`) I think it is cleaner to be consistent,
Given how common these are in test code, I'd say using static imports for all Mockito methods would be the way to go.

bvdzwet (changes requested):

osgp/platform/osgp-adapter-ws-smartmetering/src/main/java/org/opensmartgridplatform/adapter/ws/smartmetering/infra/jms/messageprocessor/DomainResponseMessageProcessor.java
rlemmers: Agree. I updated the message.

osgp/platform/osgp-adapter-ws-smartmetering/src/test/java/org/opensmartgridplatform/adapter/ws/smartmetering/infra/jms/messageprocessor/DomainResponseMessageProcessorTest.java
rlemmers: Added org.mockito.Mockito.doThrow as a static import.

bvdzwet (approved):

PR #575, opened 2021-02-16
Flex 5349/fix get status

robindenadel:


osgp/platform/osgp-adapter-ws-publiclighting/src/main/java/org/opensmartgridplatform/adapter/ws/publiclighting/endpoints/PublicLightingAdHocManagementEndpoint.java
```diff
@@ -255,6 +256,26 @@ public GetStatusResponse getGetStatusResponse(@OrganisationIdentification final
...
+        } else {
+            deviceStatus = (DeviceStatus) message.getDataObject();
+        }
+        return this.adHocManagementMapper.map(deviceStatus,
+                org.opensmartgridplatform.adapter.ws.schema.publiclighting.adhocmanagement.DeviceStatus.class);
+    }
+
+    private DeviceStatus createDeviceStatusFromLightSensorStatus(final LightSensorStatus lightSensorStatus) {
+        final List<LightValue> sensorValues = new ArrayList<>();
+        sensorValues.add(new LightValue(1, lightSensorStatus.isOn(), 100));
```
bvdzwet, line 273: The dim value does not make a lot of sense for the light sensor status. If setting it to null causes issues, some default is probably pragmatically OK for now, and 100 seems as good as any other. Something similar applies to the defaults below in the `DeviceStatus` constructor.
Thorough fixing of modelling issues with light values and device statuses looks definitely out of scope for what was considered in making get status work by fixing correlation UID issues.

osgp/platform/osgp-adapter-ws-tariffswitching/src/main/java/org/opensmartgridplatform/adapter/ws/tariffswitching/application/mapping/ws/TariffScheduleToScheduleConverter.java
```diff
@@ -40,26 +38,21 @@ public ScheduleEntry convert(final TariffSchedule source, final Type<? extends S
...
-            // schedule.setLightValue( Arrays.asList(new LightValue(null,
-            // !source.isHigh(), null)) )
-
-            final List<LightValue> lightValues = new ArrayList<>();
-            for (final TariffValue tariffValue : source.getTariffValue()) {
+        // Set the lightvalue
+        // For now a High tariff means the Relay is switched off (Situation
+        // in Zaltbommel)
+        // schedule.setLightValue( Arrays.asList(new LightValue(null,
+        // !source.isHigh(), null)) )
```
bvdzwet, line 45: I see this was only a formatting change. I wouldn't have complained if these lines would have been deleted, though.
Just noticed the switch from tariff to light value. Some comment might have been helpful. Safe to ignore this further as far as the get status changes are concerned.

Only wondering why this code was touched at all, but that might be already explained looking at the validation exception that was removed from the `LightValue` constructor.

osgp/protocol-adapter-iec60870/osgp-protocol-adapter-iec60870/src/main/java/org/opensmartgridplatform/adapter/protocol/iec60870/application/services/LightMeasurementDeviceResponseService.java
```diff
@@ -46,6 +48,9 @@
     @Autowired
     private LightMeasurementService lightMeasurementService;
 
+    @Autowired
+    private CorrelationUidPerDevice correlationUidQueuePerDevice;
```
bvdzwet, line 52: Why is there "Queue" in the name of this field? If this is a left-over from an earlier version of the code and not intentional, it is probably better to take the queue out of the name.

osgp/protocol-adapter-iec60870/osgp-protocol-adapter-iec60870/src/main/java/org/opensmartgridplatform/adapter/protocol/iec60870/application/services/LightMeasurementDeviceResponseService.java
```diff
@@ -139,17 +157,27 @@ private void sendLightSensorStatusResponse(final MeasurementGroupDto measurement
...
             final ResponseMetadata responseMetadata) {
 
-        final ResponseMetadata rm = new ResponseMetadata.Builder()
-                .withCorrelationUid(responseMetadata.getCorrelationUid())
-                .withDeviceIdentification(device.getDeviceIdentification())
+        final String deviceIdentification = device.getDeviceIdentification();
+
+        Optionals.ifPresentOrElse(this.correlationUidQueuePerDevice.dequeu(deviceIdentification),
+                uid -> this.sendLightSensorStatus(uid, lightSensorStatus, deviceIdentification, responseMetadata),
+                () -> LOGGER.info("No correlation UID found for device identification {}", deviceIdentification));
```
bvdzwet, line 172: I am unsure how to regard the case of the missing correlation UID. I would have guessed this would warrant warning, possibly error level logging. If you made a conscious decision info level is suitable here, you can leave it as is.

Further in the review I notices correlationUids being kept with the identification of the light measurement device instead of the gateway device (alternatives that were discussed earlier). Am I correct to guess that in the current set up this logging would be triggered for all but one of the light measurement devices sharing the gateway?
If so that looks like it makes things work for now, but may be good to revisit at a later time as it seems not entirely desirable.

osgp/protocol-adapter-iec60870/osgp-protocol-adapter-iec60870/src/main/java/org/opensmartgridplatform/adapter/protocol/iec60870/infra/CorrelationUidPerDevice.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.Optional;
+
+public class CorrelationUidPerDevice {
+
+    private final Map<String, LinkedList<String>> correlationUidQueuePerDevice = new HashMap<>();
+
+    public void enqueu(final String deviceIdentification, final String correlationUid) {
```
bvdzwet, line 19: Given the conventional spelling, I think `enqueue` might be a better name.

osgp/protocol-adapter-iec60870/osgp-protocol-adapter-iec60870/src/main/java/org/opensmartgridplatform/adapter/protocol/iec60870/infra/CorrelationUidPerDevice.java
```diff
@@ -0,0 +1,33 @@
...
+        this.correlationUidQueuePerDevice.computeIfAbsent(deviceIdentification, key -> new LinkedList<>())
+                .add(correlationUid);
+    }
+
+    public Optional<String> dequeu(final String deviceIdentification) {
```
bvdzwet, line 24: Given the conventional spelling, I think `dequeue` might be a better name.

osgp/protocol-adapter-iec60870/osgp-protocol-adapter-iec60870/src/main/java/org/opensmartgridplatform/adapter/protocol/iec60870/infra/CorrelationUidPerDevice.java
```diff
@@ -0,0 +1,33 @@
...
+        final String correlationUid = this.correlationUidQueuePerDevice
+                .getOrDefault(deviceIdentification, new LinkedList<>())
+                .poll();
+        if (correlationUid == null) {
+            return Optional.empty();
+        }
+        return Optional.of(correlationUid);
```
bvdzwet, line 31: This works, but you could consider the following alternative, which seems to make more stylish use of what optional has to offer:
`return Optional.ofNullable(correlationUid);`

osgp/protocol-adapter-iec60870/osgp-protocol-adapter-iec60870/src/test/java/org/opensmartgridplatform/adapter/protocol/iec60870/infra/messaging/DeviceRequestMessageListenerTest.java
```diff
@@ -41,8 +44,15 @@
     @Mock
     private DeviceResponseMessageSender deviceResponseMessageSender;
 
+    @BeforeEach
+    void setup() {
+        ReflectionTestUtils.setField(this.deviceRequestMessageListener, "correlationUidPerDevice",
+                new CorrelationUidPerDevice());
```
bvdzwet, line 50: It looks like this adds the CorreationUidPerDevice just to avoid exceptions when it is called.
Did you consider adding a mock instead and verifying in a new test that the correlation UID is enqueued for the proper device?
On the one hand this is covered by any (integration) test where the status is retrieved. On the other hand I wonder if unit testing this side, plus unit testing the class that dequeues correlation UIDs from CorrelationUidPerDevice would give some more certainty about the working of correlation UID matching when performing later changes to the code.

bvdzwet (changes requested):

bvdzwet (approved):

PR #593, opened 2021-03-25
Slim 1883/update subscription information

coendamen:


integration-tests/cucumber-tests-platform-smartmetering/src/test/resources/features/osgp-adapter-ws-smartmetering/installation/SetSubscriptionInformation.feature
```diff
@@ -0,0 +1,20 @@
...
+      | DeviceIdentification | TEST1024000000001 |
+      | IpAddress            | 10.0.0.1          |
+      | BtsId                | 20                |
+      | CellId               | 1                 |
+    Then the set subscription information response should be returned
+      | DeviceIdentification | TEST1024000000001 |
+      | Result               | OK                |
+      | IpAddress            | 10.0.0.1          |
+      | BtsId                | 20                |
+      | CellId               | 1                 |
```
kroesctrl, line 20: Numbers should be aligned to the right

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/infra/jms/ws/messageprocessors/SetSubscriptionInformationRequestMessageProcessor.java
```diff
@@ -0,0 +1,43 @@
+/**
+ * Copyright 2015 Smart Society Services B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
```
kroesctrl, line 8: 2021 Alliander

osgp/platform/osgp-domain-core/src/main/java/org/opensmartgridplatform/domain/core/valueobjects/smartmetering/SetSubscriptionInformationRequestData.java
```diff
@@ -0,0 +1,43 @@
...
+import java.io.Serializable;
+
+import lombok.Data;
+
+@Data
+public class SetSubscriptionInformationRequestData implements Serializable {
+
+    private static final long serialVersionUID = 333099329546171974L;
+
+    String meId;
```
kroesctrl, line 21: private properties?

osgp/platform/osgp-domain-core/src/main/java/org/opensmartgridplatform/domain/core/valueobjects/smartmetering/SetSubscriptionInformationResponseData.java
```diff
@@ -0,0 +1,43 @@
...
+import java.io.Serializable;
+
+import lombok.Data;
+
+@Data
+public class SetSubscriptionInformationResponseData implements Serializable {
+
+    private static final long serialVersionUID = 333099329546171974L;
+
+    String meId;
```
kroesctrl, line 21: Private properties?

kroesctrl (changes requested):

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/infra/jms/ws/messageprocessors/SetSubscriptionInformationRequestMessageProcessor.java
coendamen: done

osgp/platform/osgp-domain-core/src/main/java/org/opensmartgridplatform/domain/core/valueobjects/smartmetering/SetSubscriptionInformationRequestData.java
coendamen: of course, done

integration-tests/cucumber-tests-platform-smartmetering/src/test/resources/features/osgp-adapter-ws-smartmetering/installation/SetSubscriptionInformation.feature
coendamen: corrected, but actually, the formatter puts them on the left side.

osgp/platform/osgp-domain-core/src/main/java/org/opensmartgridplatform/domain/core/valueobjects/smartmetering/SetSubscriptionInformationResponseData.java
coendamen: done

coendamen:
retest this please

osgp/platform/osgp-adapter-ws-smartmetering/src/main/java/org/opensmartgridplatform/adapter/ws/smartmetering/endpoints/SmartMeteringInstallationEndpoint.java
```diff
@@ -447,4 +470,72 @@ public DecoupleMbusDeviceByChannelResponse getDecoupleMbusDeviceByChannelRespons
         }
         return response;
     }
-}
+
+    @PayloadRoot(localPart = "SetSubscriptionInformationRequest", namespace = SMARTMETER_INSTALLATION_NAMESPACE)
+    @ResponsePayload
+    public SetSubscriptionInformationAsyncResponse setSubscriptionInformation(
```
bvdzwet, line 476: I'd like to check whether this ended up in the intended location in the GXF code base.
Looking at the story I see: "The IP-address must be saved to the OSGP core database via the GXF management SOAP service."
The place it ended up is the Smart Metering Installation SOAP service.

Given that the data that is managed is in the core `Device` entity, and not specific to the `SmartMeter` entity you could expect that this would end up in osgp-adapter-ws-core `DeviceManagementEndpoint`.

If it was intended to be smart meter only (could be, though a bit strange with the data in the general `Device`) there is still the `SmartMeteringManagementEndpoint` that looks it would fit better with the description in the story.

osgp/platform/osgp-adapter-ws-smartmetering/src/main/java/org/opensmartgridplatform/adapter/ws/smartmetering/endpoints/SmartMeteringInstallationEndpoint.java
```diff
@@ -447,4 +470,72 @@ public DecoupleMbusDeviceByChannelResponse getDecoupleMbusDeviceByChannelRespons
...
+                    request.getDeviceIdentification(), organisationIdentification, e);
+
+            this.handleException(e);
+        }
+        return response;
+    }
+
+    @PayloadRoot(localPart = "SetSubscriptionInformationAsyncRequest", namespace = SMARTMETER_INSTALLATION_NAMESPACE)
+    @ResponsePayload
+    public SetSubscriptionInformationResponse getSetSubscriptionInformationResponse(
```
bvdzwet, line 520: I'd like to check whether "SubscriptionInformation" is a desirable name for the information that is maintained in the newly added operations.
My feeling is that it has relevant meaning in some other context than GXF, which might also explain all the unused fields (MeId, Esn, UimId, EqId. Mdn, Status, CustCode and SupplierReferenceId), where only IpAddress, BtsId and CellId are used.
Would something like "NetworkingInformation" be a better name here, or do I need to acquaint myself a bit more with the meaning of "SubscriptionInformation"?

osgp/platform/osgp-domain-core/src/main/java/org/opensmartgridplatform/domain/core/entities/Device.java
```diff
@@ -193,6 +193,12 @@
     @Enumerated(EnumType.STRING)
     private IntegrationType integrationType = IntegrationType.WEB_SERVICE;
 
+    @Column
+    private Integer cellId;
+
+    @Column
+    private Integer btsId;
```
bvdzwet, line 200: These are updated together with the `networkAddress`. Would it make sense to keep them a bit closer together in the source file?

osgp/platform/osgp-domain-core/src/main/java/org/opensmartgridplatform/domain/core/entities/SmartMeter.java
```diff
@@ -14,13 +15,13 @@
 import org.opensmartgridplatform.domain.core.valueobjects.Address;
 import org.opensmartgridplatform.domain.core.valueobjects.GpsCoordinates;
 
+import lombok.EqualsAndHashCode;
+
+@EqualsAndHashCode(callSuper = true)
```
bvdzwet, line 20: I don't think the `EqualsAndHashCode` provided by Lombok is appropriate for entities.
The equals implementation in `Device` takes into account the business key (`deviceIdentification`) which is the correct way to implement equality if such a business key is available.
Lombok adds in the other fields when building its equals implementation. A SmartMeter is not a different SmartMeter if any of these other fields have different values.

osgp/shared/osgp-dto/src/main/java/org/opensmartgridplatform/dto/valueobjects/smartmetering/SmartMeteringDeviceDto.java
```diff
@@ -11,207 +11,35 @@
...
     private String mbusManufacturerIdentification;
-
     private Short mbusVersion;
-
     private Short mbusDeviceTypeIdentification;
-
     private byte[] mbusDefaultKey;
+    private String ipAddress;
+    private int cellId;
+    private int btsId;
```
bvdzwet, line 43: It feels odd that these are included here and not with the `SmartMeteringDevice` which is tightly related.
Could this have to do with an underlying issue that the new fields belong with Device rather than with SmartMeter?

osgp/shared/osgp-ws-smartmetering/src/main/resources/SmartMeteringInstallation.wsdl
```diff
@@ -145,48 +144,48 @@
...
       </wsdl:output>
     </wsdl:operation>
     <wsdl:operation name="GetCoupleMbusDeviceByChannelResponse">
       <wsdl:input message="tns:CoupleMbusDeviceByChannelAsyncRequest"
-        name="CoupleMbusDeviceByChannelAsyncRequest">
+                  name="CoupleMbusDeviceByChannelAsyncRequest">
       </wsdl:input>
       <wsdl:output message="tns:CoupleMbusDeviceByChannelResponse"
-        name="CoupleMbusDeviceByChannelResponse">
+                   name="CoupleMbusDeviceByChannelResponse">
```
bvdzwet, line 188: It looks like you are using some other formatting settings than have been used before, leading to lots of distracting formatting updates.

bvdzwet (changes requested):
A number of classes have varying degrees of formatting issues (I noticed some in: EventServiceTest, InstallationService, SmartMeterService, SmartMeterServiceTest, DeviceInstallationEndpoint, SmartMeteringInstallationEndpoint, SmartMeteringRequestMessageSender, SetSubscriptionInformationSteps, DeviceFactory, SetSubscriptionInformationRequestFactory, SmartMeteringInstallationClient, Device).

There appear to be quite a lot of files that were modified and that have no modifications related to the story. This does not help to focus on what should actually be done. If cleanup or other changes are necessary to code that does not need to be touched, it would be better to deal with this outside the pull request for the story.

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/InstallationService.java
```diff
@@ -71,29 +71,79 @@ public InstallationService() {
...
+        responseData.setEsn(requestData.getEsn());
+        responseData.setUimId(requestData.getUimId());
+        responseData.setEqId(requestData.getEqId());
+        responseData.setIpAddress(updatedSmartMeter.getIpAddress());
+        responseData.setMdn(requestData.getMdn());
+        responseData.setBtsId(updatedSmartMeter.getBtsId());
+        responseData.setCellId(updatedSmartMeter.getCellId());
+        responseData.setStatus(requestData.getStatus());
+        responseData.setCustCode(requestData.getCustCode());
+        responseData.setSupplierReferenceId(requestData.getSupplierReferenceId());
```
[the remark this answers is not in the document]
coendamen: Yes, the scenario here is a 2-phase action
- call from SMHE recipe to get the CDMA info from zSmart.
- on success, the data should be pushed to GXF. 
- on success, the entire set of data retrieved from zSmart should be shown in SMHE after the Device was successfully updated.

This means I have to round trip all of the data because of the asynchronous character of SMHE - GXF communication.

There is no way to save the data to be shown in SMHE for only the first step and then continue with the next step. The data is saved via ResponseProcessors that act upon a Response message from GXF.

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/SmartMeterService.java
```diff
@@ -62,32 +68,78 @@ public void storeMeter(final String organisationIdentification, final AddSmartMe
...
+        if (this.smartMeterRepository.findByDeviceIdentification(deviceIdentification) != null) {
             throw new FunctionalException(FunctionalExceptionType.EXISTING_DEVICE, ComponentType.DOMAIN_SMART_METERING);
         }
+    }
+
+    private SmartMeter validateSmartMeterExists(final String deviceIdentification) throws FunctionalException {
+        final SmartMeter smartMeter = this.smartMeterRepository.findByDeviceIdentification(deviceIdentification);
+
+        if (smartMeter == null) {
+            throw new FunctionalException(FunctionalExceptionType.EXISTING_DEVICE, ComponentType.DOMAIN_SMART_METERING);
```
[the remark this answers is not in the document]
coendamen: you are correct, I changed it to match the proper exception.

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/SmartMeterService.java
```diff
@@ -62,32 +68,78 @@ public void storeMeter(final String organisationIdentification, final AddSmartMe
...
         return this.mapperFactory.getMapperFacade().map(smartMeteringDevice, SmartMeter.class);
     }
 
+    public SmartMeter updateSubscriptionInformation(final String deviceIdentification, final String ipAddress,
+            final Integer btsId, final Integer cellId) throws FunctionalException {
+
+        SmartMeter smartMeter = this.validateSmartMeterExists(deviceIdentification);
+        boolean subscriptionInformationChanged = false;
+
+        if (ipAddress != null) {
```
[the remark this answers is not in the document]
coendamen: the problem here is that the zSmart interface, where we get this data, can return a value, and empty string or null. In that case, when a value is null/empty, we do not want to update any existing value. Therefore the null check. If it was already null, it means it was never set in the database.

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/SmartMeterService.java
```diff
@@ -62,32 +68,78 @@ public void storeMeter(final String organisationIdentification, final AddSmartMe
...
+            } catch (final UnknownHostException e) {
+                log.error("Invalid ip address found {} for device {}", ipAddress, deviceIdentification);
+                throw new FunctionalException(FunctionalExceptionType.INVALID_IP_ADDRESS,
+                        ComponentType.DOMAIN_SMART_METERING);
+            }
+        }
+
+        if (btsId != null) {
+            smartMeter.setBtsId(btsId);
+            subscriptionInformationChanged = true;
```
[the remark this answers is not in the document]
coendamen: yes, I agree this could be refactored into a separate method.

osgp/platform/osgp-adapter-ws-smartmetering/src/main/java/org/opensmartgridplatform/adapter/ws/smartmetering/endpoints/SmartMeteringInstallationEndpoint.java
coendamen: after talking with our PO, the action was part of the installation process. so it is added there. I will recheck if this is the accepted location.

osgp/platform/osgp-adapter-ws-smartmetering/src/main/java/org/opensmartgridplatform/adapter/ws/smartmetering/endpoints/SmartMeteringInstallationEndpoint.java
coendamen: this will be changed to SetCommunicationNetworkInformation

osgp/platform/osgp-domain-core/src/main/java/org/opensmartgridplatform/domain/core/entities/Device.java
coendamen: agree.

osgp/shared/osgp-dto/src/main/java/org/opensmartgridplatform/dto/valueobjects/smartmetering/SmartMeteringDeviceDto.java
coendamen: I double checked this during development with our PO, should be part of Device (i.e. any CDMA device).

osgp/shared/osgp-ws-smartmetering/src/main/resources/SmartMeteringInstallation.wsdl
coendamen: yes, the formatter is giving us headaches, actually I formatted all files based on our last version (from last week) .....

osgp/platform/osgp-domain-core/src/main/java/org/opensmartgridplatform/domain/core/entities/SmartMeter.java
coendamen: you are correct. good to know.

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/SmartMeterService.java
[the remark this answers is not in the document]
coendamen: I removed the subscriptionInformationChanged which was added to track any change, But because this service is called in the first place, there is a change anyway, so the boolean check is unnecessary.

integration-tests/cucumber-tests-platform-smartmetering/src/test/java/org/opensmartgridplatform/cucumber/platform/smartmetering/glue/steps/ws/smartmetering/smartmeteringinstallation/SetSubscriptionInformationSteps.java
```diff
@@ -0,0 +1,72 @@
...
+                        .fromScenarioContext(extendedParameters);
+        final SetSubscriptionInformationResponse response =
+                this.smartMeteringInstallationClient.getSetSubscriptionInformationResponse(
+                        setSubscriptionInformationAsyncRequest);
+
+        assertThat(response.getResult()).as("Result").isNotNull();
+        assertThat(response.getResult().name()).as("Result").isEqualTo(responseSettings.get("Result"));
+        assertThat(response.getIpAddress()).isEqualTo(responseSettings.get("IpAddress"));
+        assertThat(response.getBtsId()).isEqualTo(Integer.parseInt(responseSettings.get("BtsId")));
+        assertThat(response.getCellId()).isEqualTo(Integer.parseInt(responseSettings.get("CellId")));
```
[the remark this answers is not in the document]
coendamen: I moved them to the PlatFormKeys class.

integration-tests/cucumber-tests-platform-smartmetering/src/test/java/org/opensmartgridplatform/cucumber/platform/smartmetering/support/ws/smartmetering/installation/SetSubscriptionInformationRequestFactory.java
```diff
@@ -0,0 +1,39 @@
...
+import org.opensmartgridplatform.cucumber.platform.smartmetering.support.ws.smartmetering.RequestFactoryHelper;
+
+public class SetSubscriptionInformationRequestFactory {
+
+    public static SetSubscriptionInformationRequest fromParameters(final Map<String, String> inputSettings) {
+        final SetSubscriptionInformationRequest request = new SetSubscriptionInformationRequest();
+        request.setDeviceIdentification(inputSettings.get("DeviceIdentification"));
+        request.setIpAddress(inputSettings.get("IpAddress"));
+        request.setBtsId(Integer.parseInt(inputSettings.get("BtsId")));
+        request.setCellId(Integer.parseInt(inputSettings.get("CellId")));
```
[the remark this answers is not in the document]
coendamen: done

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/InstallationService.java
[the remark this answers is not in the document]
coendamen: Gerbert agreed to push and return only the relevant values.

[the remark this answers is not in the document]
bvdzwet: Based on my comment, and your final reaction about only pushing and returning relevant values, I would expect most of these lines to have been removed.
Apparently, at least `SetCommunicationNetworkInformationRequestData` and `SetCommunicationNetworkInformationResponseData` still have fields that I no longer expect to see (all those echoed back from `requestData` instead of from the `updatedSmartMeter`).

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/SmartMeterService.java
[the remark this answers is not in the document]
bvdzwet: Whatever interfaces applications on top of GXF use outside GXF itself should not be the direct reason for how something is built in the interface from applications to GXF.
That said, I think in itself the logic where only those fields that are provided in the request are used to update the device and there is no way to remove IP address, BTS ID and Cell ID, other than setting them to some non-emtpy "out-of-use" kind of defaults could work.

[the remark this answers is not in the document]
kroesctrl: So if they say, we don't know it anymore (null), then we're not going to update our value?

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/SmartMeterService.java
```diff
@@ -62,32 +68,72 @@ public void storeMeter(final String organisationIdentification, final AddSmartMe
...
+        }
+
+        if (btsId != null) {
+            smartMeter.setBtsId(btsId);
+        }
+
+        if (cellId != null) {
+            smartMeter.setCellId(cellId);
+        }
+
```
kroesctrl, line 125: Same cases here, values cannot be set back to null.......

osgp/platform/osgp-adapter-domain-smartmetering/src/test/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/SmartMeterServiceTest.java
```diff
@@ -0,0 +1,192 @@
...
+        final SmartMeter smartMeter = new SmartMeter();
+
+        when(this.protocolInfoRepository.findByProtocolAndProtocolVersion(any(), any())).thenReturn(null);
+
+        try {
+            this.smartMeterService.storeMeter(organisationIdentification, addSmartMeterRequest, smartMeter);
+            fail("Should throw FunctionalException");
+        } catch (final FunctionalException e) {
+            assertThat(e.getExceptionType()).isEqualTo(FunctionalExceptionType.UNKNOWN_PROTOCOL_NAME_OR_VERSION);
+        }
```
kroesctrl, line 98: Code suggestion:
```
Exception exception = assertThrows(FunctionalException.class, () -> {
         this.smartMeterService.storeMeter(organisationIdentification, addSmartMeterRequest, smartMeter);
}); 
   assertThat(exception.getExceptionType()).isEqualTo(FunctionalExceptionType.UNKNOWN_PROTOCOL_NAME_OR_VERSION);
```

osgp/platform/osgp-adapter-domain-smartmetering/src/test/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/SmartMeterServiceTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+        when(this.smartMeterRepository.findByDeviceIdentification(deviceIdentification)).thenReturn(null);
+
+        try {
+            this.smartMeterService.updateCommunicationNetworkInformation(deviceIdentification, ipAddress, btsId,
+                    cellId);
+            fail("Should throw FunctionalException");
+        } catch (final FunctionalException e) {
+            assertThat(e.getExceptionType()).isEqualTo(FunctionalExceptionType.UNKNOWN_DEVICE);
+        }
```
kroesctrl, line 146: assertThrows

osgp/platform/osgp-adapter-domain-smartmetering/src/test/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/SmartMeterServiceTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+        when(this.smartMeterRepository.findByDeviceIdentification(deviceIdentification)).thenReturn(new SmartMeter());
+
+        try {
+            this.smartMeterService.updateCommunicationNetworkInformation(deviceIdentification, ipAddress, btsId,
+                    cellId);
+            fail("Should throw FunctionalException");
+        } catch (final FunctionalException e) {
+            assertThat(e.getExceptionType()).isEqualTo(FunctionalExceptionType.INVALID_IP_ADDRESS);
+        }
```
kroesctrl, line 165: assertThrows

kroesctrl (changes requested):

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/InstallationService.java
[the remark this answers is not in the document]
coendamen: you are right, I missed those.

osgp/platform/osgp-adapter-ws-smartmetering/src/main/java/org/opensmartgridplatform/adapter/ws/smartmetering/endpoints/SmartMeteringInstallationEndpoint.java
bvdzwet: While the class names have been renamed, there still is something remaining regarding the Subscription Information (the `exceptionContext` input `String` for the exception thrown if the result from setting the Communication Network Information is not OK).

kroesctrl (approved):

osgp/platform/osgp-adapter-ws-smartmetering/src/main/java/org/opensmartgridplatform/adapter/ws/smartmetering/endpoints/SmartMeteringInstallationEndpoint.java
coendamen: > While the class and method names etc. have been renamed, there still is something remaining regarding the Subscription Information (the `exceptionContext` input `String` for the exception thrown if the result from setting the Communication Network Information is not OK).

fixed.

osgp/shared/osgp-dto/src/main/java/org/opensmartgridplatform/dto/valueobjects/smartmetering/SmartMeteringDeviceDto.java
bvdzwet: I think this is due to the fact that the hierarchy from the entities is not reflected in the value objects.
This way value objects and DTOs for specific sub-types of `Device` would need to repeat the generic device information in their fields. Not sure, but it may be that it could be a good thing to prevent this repetition of generic device fields by having the same hierarchy in the value objects and DTOs. Given this is not how the code was organized this would **not** be something to be addressed in scope of this pull request I think.
To avoid misunderstanding, not every `Device` is a CDMA device, devices can be connected with other communication methods as well.

osgp/platform/osgp-adapter-ws-smartmetering/src/main/java/org/opensmartgridplatform/adapter/ws/smartmetering/application/services/InstallationService.java
```diff
@@ -227,14 +224,35 @@ public String enqueueDecoupleMbusDeviceByChannelRequest(@Identification final St
...
         return correlationUid;
     }
 
-}
+    public String enqueueSetCommunicationNetworkInformationRequest(
+            @Identification final String organisationIdentification, @Identification final String deviceIdentification,
+            final SetCommunicationNetworkInformationRequestData setCommunicationNetworkInformationRequestData,
+            final int messagePriority, final Long scheduleTime) {
+
+        log.debug("enqueueSetSCommunicationNetworkRequest called with organisation {} and device {}",
```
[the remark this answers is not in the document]
coendamen: I'm afraid it is.

osgp/platform/osgp-adapter-ws-smartmetering/src/main/java/org/opensmartgridplatform/adapter/ws/smartmetering/endpoints/SmartMeteringInstallationEndpoint.java
```diff
@@ -58,7 +64,8 @@
 @Endpoint
 public class SmartMeteringInstallationEndpoint extends SmartMeteringEndpoint {
 
-    private static final String SMARTMETER_INSTALLATION_NAMESPACE = "http://www.opensmartgridplatform.org/schemas/smartmetering/sm-installation/2014/10";
+    private static final String SMARTMETER_INSTALLATION_NAMESPACE =
+            "http://www.opensmartgridplatform" + ".org/schemas/smartmetering/sm-installation/2014/10";
```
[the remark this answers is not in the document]
coendamen: indeed, fixed.

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/infra/jms/MessageMetadata.java
```diff
@@ -173,6 +181,8 @@ public String toString() {
                 ", domain=" + this.domain +
                 ", domainVersion=" + this.domainVersion +
                 ", ipAddress=" + this.ipAddress +
+                ", cellId=" + this.cellId +
+                ", btsId=" + this.btsId +
```
bvdzwet, line 185: Did not pay attention to this the first review, but I noticed it now.
The infra JMS MessageMetadata was set up for metadata with requests to be passed as JMS message properties.
These were values important to processing requests that were not part of the content of the request types.
Scheduling and retry properties, for instance are passes through SOAP headers, and the IP address is used because in some cases core can decide to put in the IP address of the device with the request or its gateway device.

As far as I understand the Cell ID and BTS ID are pieces of information that are passed in as part of the `SetCommunicationNetworkInformationRequest`. The IP address is in the request as well, but for setting the communication network information this is as part of the data, and not as part of the information passed by core to the protocol adapter through JMS metadata as described earlier in this comment.

bvdzwet (changes requested):

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/infra/jms/MessageMetadata.java
coendamen: yes, this was added because it will be needed by the protocol adapter(s) to throttle/limit traffic to "stations/cells".

I assume this was discussed during the inter team refinements? 

So this PR just pushes the data to GXF, so it can later be sent as metadata to the protocol adapter where the "traffic manager" can decide whether to allow/queue the request based on the bts/cell.

PR #604, opened 2021-04-07
OC-718 ~ Default ping size to 1

Marcel-Jansen:
device.ping.size=0 is not always resulting in an echo reply.
Therefore setting the default to device.ping.size=1


bvdzwet (approved):

bvdzwet:
retest this please

PR #609, opened 2021-04-14
FDP-36 ~ Adds light sensor status to GetStatusResponse

smvdheijden:


osgp/shared/osgp-ws-publiclighting/src/main/resources/schemas/pl-adhocmanagement.xsd
```diff
@@ -102,7 +102,10 @@
         <xsd:element name="Result" type="common:OsgpResultType"
           minOccurs="1" />
         <xsd:element name="Description" type="xsd:string" />
-        <xsd:element name="DeviceStatus" type="tns:DeviceStatus" />
+        <xsd:choice>
+            <xsd:element name="DeviceStatus" type="tns:DeviceStatus" />
+            <xsd:element name="LightSensorStatus" type="tns:LightSensorStatus" />
```
bvdzwet, line 107: Indented by 4 instead of 2 spaces.

The other thing I was wondering about is, what the difference would be if this were a `Status` (new type) instead of the choice. Then `DeviceStatus` and `LightSensorStatus` could extend `Status`. Do you know if this would result in nicer code or not?

osgp/shared/osgp-ws-publiclighting/src/main/resources/schemas/pl-adhocmanagement.xsd
```diff
@@ -287,6 +281,17 @@
...
   </xsd:complexType>
 
+  <xsd:complexType name="LightSensorStatus">
+    <xsd:annotation>
+      <xsd:documentation>
+        Light sensor status represents the retrieved status of a light measurement device.
+      </xsd:documentation>
+    </xsd:annotation>
+    <xsd:sequence>
+      <xsd:element name="Dark" type="xsd:boolean" maxOccurs="1" />
```
bvdzwet, line 291: A specific type for `LightSensorStatusValue` (or another name) that has enumerated values for the sensor status would look a bit nicer I think. Then it would be immediately clear that the possible values are `Dark` or `Light` (I would imagine) and the meaning would not have to be derived from a boolean.

bvdzwet (changes requested):

osgp/protocol-adapter-iec61850/osgp-protocol-adapter-iec61850/src/main/java/org/opensmartgridplatform/adapter/protocol/iec61850/infra/networking/services/Iec61850DeviceConnectionService.java
```diff
@@ -68,6 +67,8 @@
 
   @Autowired private Iec61850Client iec61850Client;
 
+  @Autowired private int iec61850DefaultPort;
```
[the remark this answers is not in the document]
smvdheijden: I had not yet considered this. Although I agree with you that this could be an improvement, for now I'll leave it as is.

osgp/shared/osgp-ws-publiclighting/src/main/resources/schemas/pl-adhocmanagement.xsd
```diff
@@ -287,6 +259,46 @@
     </xsd:sequence>
   </xsd:complexType>
 
+  <xsd:complexType name="Status" />
```
bvdzwet, line 262: If you change this to `<xsd:complexType name="Status" abstract="true" />` the generated class for `Status` will be abstract, which appears to be a better fit with the interface for the related domain objects.

osgp/shared/osgp-ws-publiclighting/src/main/resources/schemas/pl-adhocmanagement.xsd
```diff
@@ -287,6 +259,46 @@
...
+  <xsd:complexType name="LightSensorStatus">
+    <xsd:annotation>
+      <xsd:documentation>
+        Light sensor status represents the retrieved status of a light measurement device.
+      </xsd:documentation>
+    </xsd:annotation>
+    <xsd:complexContent>
+      <xsd:extension base="tns:Status">
+        <xsd:sequence>
+          <xsd:element name="Status" type="tns:LightSensorStatusType" maxOccurs="1" />
```
bvdzwet, line 296: Attribute `maxOccurs` has its default value, like `minOccurs` that is left out here.
Probably best to include neither.

bvdzwet (changes requested):

osgp/platform/osgp-adapter-domain-publiclighting/src/main/java/org/opensmartgridplatform/adapter/domain/publiclighting/application/services/AdHocManagementService.java
```diff
@@ -126,7 +128,9 @@ public void handleGetLightSensorStatusResponse(
     response.setOsgpException(exception);
     response.setResult(deviceResult);
     if (lightSensorStatusDto != null) {
-      response.setLightSensorStatus(new LightSensorStatus(lightSensorStatusDto.isOn()));
+      response.setLightSensorStatus(
+          new LightSensorStatus(
+              LightSensorStatusType.valueOf(lightSensorStatusDto.getStatus().toString())));
```
bvdzwet, line 133: When used in a functional way (and not only informational) it is probably better not to rely on `toString()` but use something more specific. For enums I think either `name()` or a custom property would be more robust choices.

bvdzwet (approved):

PR #616, opened 2021-04-22
Feature/fdp 37 get status test for lmd

smvdheijden:


integration-tests/cucumber-tests-protocol-iec60870/src/test/java/org/opensmartgridplatform/cucumber/protocol/iec60870/glue/steps/ProcessImageSteps.java
```diff
@@ -0,0 +1,100 @@
...
+        entry, ProtocolIec60870Keys.INFORMATION_OBJECT_TYPE, Iec60870InformationObjectType.class);
+  }
+
+  private static Object informationElementValue(final Map<String, String> entry) {
+    return informationElementValue(entry.get(ProtocolIec60870Keys.INFORMATION_ELEMENT_VALUE));
+  }
+
+  private static Object informationElementValue(final String value) {
+    final String profile = (String) ScenarioContext.current().get(PROFILE);
+    if ("light_measurement_device".equals(profile)) {
```
bvdzwet, line 95: Assuming this is the same profile as specified in the scenario hooks, it might be a good idea to use a constant for this profile identifier.

This way of dealing with the values does not look really future proof, or will all profiles always have values of the same type for all information objects? Might be good enough for now though, as I don't think it would be a trivial fix, but we may need to keep this in mind with future changes.

osgp/protocol-adapter-iec60870/osgp-iec60870/src/main/java/org/opensmartgridplatform/iec60870/Iec60870Server.java
```diff
@@ -74,6 +74,27 @@ public void setProcessImage(final Map<Integer, InformationElement[][]> processIm
...
+   * Initializes the value for a specific information object address
+   *
+   * <p>Contrary to the {@link Iec60870Server#updateInformationObject(int,
+   * Iec60870InformationObjectType, Object)} method, no event is sent to the controlling station.
+   *
+   * @param informationObjectAddress
+   * @param informationObjectType
+   * @param value
+   */
+  public void initInformationObject(
```
bvdzwet, line 87: Does it make sense to initialize the information objects separately (could be if there is only one in the image)?
Otherwise it would be better to replace this by, or ad an initProcessImage method that takes the `Iec60870ProcessImage` (may be what setProcessImage already provides). Initialization gives the impression that it has to do with set up which for me stronger relates to the entire image than to the separate parts.
Creating the image in the ProcessImageSteps, and then not setting it on the server, but initializing the seperate information objects where they will together form the image again feels like taking a detour. Or is there a reason to take this seemingly longer road?

bvdzwet (changes requested):

integration-tests/cucumber-tests-protocol-iec60870/src/test/java/org/opensmartgridplatform/cucumber/protocol/iec60870/glue/steps/ProcessImageSteps.java
smvdheijden: Created Iec60870ProfileType enum.

bvdzwet (approved):

PR #625, opened 2021-05-07
Added logging

kroesctrl:


osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/infra/jms/core/OsgpCoreResponseMessageProcessor.java
```diff
@@ -180,7 +181,6 @@ protected abstract void handleMessage(
    * Suppressing S1130: Remove the declaration of thrown exception "FunctionalException"
    * FunctionalException is thrown from the implementation in subclass BundleResponseMessageProcessor.
    */
-  @SuppressWarnings("squid:S1130")
```
bvdzwet, line 183: Do not remove this SuppressWarning.
It suppresses precisely the warning that pops-up again by SonarQube, and explains why.
Even if it would be good to remove it (which I don't think) then the accompanying comment would be orphaned.
I guess this is some sort of automated clean up that is overly aggressive.
In another pull request something similar happened with another suppression (about unchecked casts).

bvdzwet (changes requested):

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/infra/jms/core/OsgpCoreResponseMessageProcessor.java
kroesctrl: I re-added die annotation, and made a change in the configuration to disable this function in the IDE

JelleHoffman (approved):

bvdzwet (approved):

PR #627, opened 2021-05-10
Clear quartz scheduler on startup and not on shutdown, this causes is…

kroesctrl:
The clear method on the scheduler, causes the triggers to be deleted in quartz.
When clearing the scheduler on shutdown, this causes quartz triggers to be removed when the application is stopped.
In kubernetes, the Rolling Update strategy first starts new instances, and the removes instances.
And because removing of an instances deletes the triggers, they are gone in the database, until a new instance is started.

This changes clears the triggers on startup, and then creates the new triggers in the scheduler based on the property files

Clearing of the triggers is required because if a trigger is removed in the application, you also want to deleted it from the database.

### DIFF
--- a/osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/application/scheduling/OsgpScheduler.java
+++ b/osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/application/scheduling/OsgpScheduler.java
@@ -44,6 +44,10 @@ public class OsgpScheduler {
   public OsgpScheduler(final Scheduler quartzScheduler) throws SchedulerException {
     this.quartzScheduler = quartzScheduler;
     LOGGER.info("Starting {}.", quartzScheduler.getSchedulerName());
+    // Clear existing jobs, to make sure the latest version is loaded
+    // and deleted triggers are removed from the database
+    this.quartzScheduler.clear();
+    LOGGER.info("Existing scheduler cleared for scheduler {}.", quartzScheduler.getSchedulerName());
   }
 
   /**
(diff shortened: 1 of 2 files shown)


osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/application/scheduling/OsgpScheduler.java
```diff
@@ -44,6 +44,10 @@
   public OsgpScheduler(final Scheduler quartzScheduler) throws SchedulerException {
     this.quartzScheduler = quartzScheduler;
     LOGGER.info("Starting {}.", quartzScheduler.getSchedulerName());
+    // Clear existing jobs, to make sure the latest version is loaded
+    // and deleted triggers are removed from the database
+    this.quartzScheduler.clear();
```
coendamen, line 49: amazing! dat we dit als team niet eerder hadden ontdekt.

osgp/shared/shared/src/test/java/org/opensmartgridplatform/shared/application/scheduling/OsgpSchedulerTest.java
```diff
@@ -0,0 +1,46 @@
...
+  @Test
+  void testClearOnConstructor() throws SchedulerException {
+    reset(this.quartzScheduler);
+    this.osgpScheduler = new OsgpScheduler(this.quartzScheduler);
+    verify(this.quartzScheduler).clear();
+  }
+
+  @Test
+  void testShutdown() throws SchedulerException {
+    reset(this.quartzScheduler);
```
coendamen, line 40: isn't reset implicitly done for a mock at the beginning of a test ?

PR #630, opened 2021-05-12
FDP-249: Fixed Sonar issues

sanderv:


bvdzwet (changes requested):
Changes generally look fine.
Some minor things could be improved, like not combining assertj isNotNull and isNotEmpty (as the latter includes the former). Commenting about setup call and verification in tests is not something we have a rule about, but my personal view is that this is not really stylish, especially not when the test is short and its structure is already clear.
Another thing not flagged by Sonar that may be an improvement is to access constants through the class they are defined in instead of through a subclass (e.g. PlatformKeys instead of PlatformPubliclightingKeys).

bvdzwet (approved):

PR #644, opened 2021-06-09
SLIM-2987-HSTN-an-uploaded-FW-file-in-SMHE-must-also-be-send-to-GXF

harrymiddelburgetall:
The existing implementation of two endpoints in FirmwareManagement are not compliant with the datamodel of the Core database. These endpoints (AddFirmwareRequest and ChangeFirmwareRequest) differ from this datamodel on two major points

- The Identification column was added as mandatory, unique and user-defined column. 
The current implementation uses either the Id column (internal database key) or a combination of Firmware's filename and related devicemodel's code to uniquely identifythe Firmware

- The current implementation assumes an one-on-one relationbetween Firmware and DeviceModel. The datamodel in the Core database has a many-to-many relation between Firmware and DeviceModel.

SMHE needs an endpoint to synchronize between it's own database and the database on the GXF platform.

- Therefor and extra endpoint was implemented: AddorChangeFirmware.
- This new endpoint consumes this new AddorChangeFirmwareRequest containing the existing Firmware class. This class originally had one ModelCode and one Manufacturer. This was replaced by a list of DeviceModels (an already existing object) containig both ModelCode and Manufacturer).
- The Firmware class now also has an Identification
- To be compliant with this modified Firmware class the AddFirmwareRequest and ChangeFirmwareRequest handling was slightly modified assuming the list of DeviceModels always contains exactly one DeviceModel.
- The Firmware class in AddFirmwareRequest and ChangeFirmwareRequest has no value for Identification but generates one serverside in case of a new Firmware.

### DIFF
--- a/osgp/platform/osgp-adapter-ws-core/src/main/java/org/opensmartgridplatform/adapter/ws/core/endpoints/FirmwareManagementEndpoint.java
+++ b/osgp/platform/osgp-adapter-ws-core/src/main/java/org/opensmartgridplatform/adapter/ws/core/endpoints/FirmwareManagementEndpoint.java
@@ -1130,14 +1130,18 @@ public class FirmwareManagementEndpoint {
 
   private String getManufacturerFromFirmware(final Firmware firmware) {
     return firmware.getDeviceModels().stream()
-        .map(dm -> dm.getManufacturer())
+        .map(
+            org.opensmartgridplatform.adapter.ws.schema.core.firmwaremanagement.DeviceModel
+                ::getManufacturer)
         .findFirst()
         .orElse(null);
   }
 
   private String getModelCodeFromFirmware(final Firmware firmware) {
     return firmware.getDeviceModels().stream()
-        .map(dm -> dm.getModelCode())
+        .map(
+            org.opensmartgridplatform.adapter.ws.schema.core.firmwaremanagement.DeviceModel
+                ::getModelCode)
         .findFirst()
         .orElse(null);
   }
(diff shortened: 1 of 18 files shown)


kevinsmeets:
retest this please

osgp/shared/osgp-ws-core/src/main/resources/schemas/firmwaremanagement.xsd
```diff
@@ -465,23 +483,11 @@
...
-        </xsd:simpleType>
-      </xsd:element>
-      <xsd:element name="ModelCode" minOccurs="0">
-        <xsd:simpleType>
-          <xsd:restriction base="xsd:string">
-            <xsd:maxLength value="255" />
-          </xsd:restriction>
-        </xsd:simpleType>
-      </xsd:element>
+       <xsd:element minOccurs="0" maxOccurs="unbounded" name="DeviceModels" type="tns:DeviceModel" />
```
kevinsmeets, line 490: Formatting: one space too many.

osgp/platform/osgp-adapter-ws-core/src/main/java/org/opensmartgridplatform/adapter/ws/core/endpoints/FirmwareManagementEndpoint.java
```diff
@@ -923,13 +937,24 @@ public ChangeFirmwareResponse changeFirmware(
...
+    // The ChangeFirmwareRequest accepts multiple DeviceModels to be related to a Firmware.
+    // This FirmwareManagementService only accepts ONE for now
+    final List<org.opensmartgridplatform.adapter.ws.schema.core.firmwaremanagement.DeviceModel>
+        wsDeviceModels = request.getFirmware().getDeviceModels();
+    String modelCode = null;
+    String manufacturer = null;
+    if (!wsDeviceModels.isEmpty()) {
+      modelCode = wsDeviceModels.get(0).getModelCode();
+      manufacturer = wsDeviceModels.get(0).getManufacturer();
+    }
```
kevinsmeets, line 949: These lines of code are the same as the lines 874 - 884. Consider creating a function.

harrymiddelburgetall: Extracted method for these lines.

kevinsmeets (approved):

PR #652, opened 2021-06-22
FDP-114 Synchronize SMR 5 invocation counter without retries

bvdzwet:
Avoids resetting the invocation counter and throwing an exception to
deal with the invocation counter initialization utilizing the GXF retry
scheduling mechanism.

This should work assuming that a connection to the management client of
the DLMS device can be set up immediately after gracefully disconnecting
from the public client in case the public client connection is utilized
to retrieve an up to date value of the invocation counter.

Resetting the invocation counter (by setting it to null in the database)
is no longer needed as the connection will be retried with an explicit
flag to initialize the invocation counter that bypasses the check in the
database on the second attempt.


bvdzwet:
retest this please

smvdheijden (approved):

harrymiddelburgetall (approved):
No comment on code. 
Tested on Bundled requests with and without bypassRetry.
Results are as expected

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/factories/DlmsConnectionHelper.java
```diff
@@ -27,20 +29,27 @@
...
   private final DevicePingConfig devicePingConfig;
+  private final Duration delayBetweenDlmsConnections;
 
   @Autowired
   public DlmsConnectionHelper(
       final InvocationCounterManager invocationCounterManager,
       final DlmsConnectionFactory connectionFactory,
-      final DevicePingConfig devicePingConfig) {
+      final DevicePingConfig devicePingConfig,
+      @Value("${dlms.connections.delay.seconds:30}") final int secondsBetweenDlmsConnections) {
```
carlogeenen, line 44: default value already present in properties file.

PR #668, opened 2021-07-12
FDP-271 ~ Uses single broker for MQTT devices.

kevinsmeets:
[@DistributionAutomation]

Signed-off-by: Kevin Smeets <kevin.smeets@cgi.com>

### DIFF
--- a/osgp/protocol-adapter-mqtt/osgp-protocol-adapter-mqtt/src/main/java/org/opensmartgridplatform/adapter/protocol/mqtt/application/services/MqttClient.java
+++ b/osgp/protocol-adapter-mqtt/osgp-protocol-adapter-mqtt/src/main/java/org/opensmartgridplatform/adapter/protocol/mqtt/application/services/MqttClient.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2021 Alliander N.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.opensmartgridplatform.adapter.protocol.mqtt.application.services;
+
+import com.hivemq.client.mqtt.MqttClientSslConfig;
+import com.hivemq.client.mqtt.mqtt3.Mqtt3AsyncClient;
+import com.hivemq.client.mqtt.mqtt3.Mqtt3Client;
+import com.hivemq.client.mqtt.mqtt3.message.connect.connack.Mqtt3ConnAck;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MqttClient {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(MqttClient.class);
+
+  private Mqtt3AsyncClient client;
+
+  public Mqtt3AsyncClient connect(
+      final String host, final int port, final MqttClientSslConfig mqttClientSslConfig) {
+    final String id = UUID.randomUUID().toString();
+    this.client =
+        Mqtt3Client.builder()
+            .identifier(id)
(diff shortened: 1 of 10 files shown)


smvdheijden (changes requested):
I believe inbound GetDataRequest messages (via InboundOsgpCoreGetDataRequestMessageProcessor) will potentially still lead to setting up a new connection with a broker? 
I would have expected to see changes in the application services of the protocol adapter for  splitting up connecting to a broker and subscribing to topics for devices. 
A possible solution would probably contain:
* MqttClient bean in the Mqtt Config, connecting to the broker at startup of the protocol adapter
* MqttDevice with unique topic(s) per device and without IP address and port (as these are only used by the MqttClient for connecting to the broker)
* For the default (unique) topic(s) per device probably a naming convention should be agreed upon. It will probably consist of a default (configurable) prefix/suffix and the device identification (which will likely be the name of the substation?)
* When a GetDataRequest is received for a device (which will be the result of communication restore mechanism in DA domain adapter) the device will subscribe to it's topic(s) using the existing MqttClient.

Could you make sure the sonar issues are resolved as well?

osgp/protocol-adapter-mqtt/osgp-protocol-adapter-mqtt/src/main/java/org/opensmartgridplatform/adapter/protocol/mqtt/application/services/SubscriptionService.java
```diff
@@ -47,11 +51,20 @@ public SubscriptionService(
     this.mqttClientDefaults = mqttClientDefaults;
   }
 
+  public void setMqttClient(final MqttClient mqttClient) {
```
smvdheijden, line 54: It seems this method is only added for testing purposes? 
Have you considered removing this method here and using `ReflectionTestUtils` to set the mqttClient in the test?

osgp/protocol-adapter-mqtt/osgp-protocol-adapter-mqtt/src/main/java/org/opensmartgridplatform/adapter/protocol/mqtt/application/services/MqttClient.java
```diff
@@ -0,0 +1,74 @@
...
+
+import com.hivemq.client.mqtt.MqttClientSslConfig;
+import com.hivemq.client.mqtt.mqtt3.Mqtt3AsyncClient;
+import com.hivemq.client.mqtt.mqtt3.Mqtt3Client;
+import com.hivemq.client.mqtt.mqtt3.message.connect.connack.Mqtt3ConnAck;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MqttClient {
```
smvdheijden, line 20: There seems to be a some similarity between this class and the MqttClientAdapter class, which makes me think it could probably use some refactoring.For now I guess you can leave it as is...

osgp/protocol-adapter-mqtt/osgp-protocol-adapter-mqtt/src/main/java/org/opensmartgridplatform/adapter/protocol/mqtt/application/services/SubscriptionService.java
kevinsmeets: Updated the unit test.

osgp/protocol-adapter-mqtt/osgp-protocol-adapter-mqtt/src/main/java/org/opensmartgridplatform/adapter/protocol/mqtt/application/services/MqttClient.java
kevinsmeets: OK

smvdheijden (approved):

PR #682, opened 2021-07-23
SLIM-2759 Update get keys service

stefanermens:


Marcel-Jansen (approved):

bvdzwet (approved):
Approved for now, because the changes result in the short term in proper configured key pairs to be used in the test environments to make things work without using different key pairs with identical keys.

As a final long term result we will need to look into the configuration of the RSA encryption classes and make sure they are applied with configuration that has logical names appropriate for the encryption scenario.

PR #684, opened 2021-07-28
Fix Maven build: added parent to module osgp. Profile 'gxf' shouldn't be …

sanderv:
- give 'osgp' a parent pom
- Don't use profile 'gxf' within gxf itself

### DIFF
--- a/osgp/pom.xml
+++ b/osgp/pom.xml
@@ -13,9 +13,7 @@
     <relativePath>../super/pom.xml</relativePath>
   </parent>
 
-  <groupId>org.osgp</groupId>
   <artifactId>osgp</artifactId>
-  <version>5.15.0-SNAPSHOT</version>
   <packaging>pom</packaging>
 
   <modules>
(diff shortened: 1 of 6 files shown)


osgp/pom.xml
```diff
@@ -6,6 +6,13 @@
   xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
+  <parent>
+    <groupId>org.opensmartgridplatform</groupId>
+    <artifactId>super-osgp</artifactId>
+    <version>5.15.0-SNAPSHOT</version>
```
bvdzwet, line 12: Now that the version is specified for the parent, can (should?) it not be removed from this POM (line 18)?
I was looking for the groupId as another candidate to be removed as it would match the groupId from the parent POM.
Then I noticed in this POM the groupId is 'org.osgp'. Is this intentionally different from the groupId used elsewhere in the platform?

bvdzwet (changes requested):
Left some comments where the parent was introduced that are probably good to take a look at.

I wonder something else regarding the missing parent. Shouldn't it be the case that all of the POMs for the open-smart-grid-platform should have a relationship through one or more parents to the super/pom? This way all dependency and plugin management from the super POM would be in place everywhere. I noticed for instance the root POM that does not appear to have this relation.
If this is indeed something to improve on, maybe not in scope of fixing the dependency check builds.

bvdzwet (approved):

PR #691, opened 2021-08-03
FDP-111 Add max schedule time for bundle requests

bvdzwet:
Nightly: https://ci.opensmartgridplatform.org/view/Nightly/job/OSGP_open-smart-grid-platform_nightly/1342/


osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/infra/jms/MessageMetadata.java
```diff
@@ -81,60 +88,51 @@ public static final MessageMetadata fromMessage(final Message message) throws JM
...
+    /*
+     * Not setting the jmsxDeliveryCount as int property named Constants.DELIVERY_COUNT, because
+     * this is not some metadata to be transferred over to new JMS messages. This delivery count is
+     * incremented by the message queue environment when a message is re-delivered to a consumer.
+     */
   }
 
-  public int getJmsxDeliveryCount() {
-    return this.jmsxDeliveryCount;
+  public static MessageMetadata.Builder newBuilder() {
```
sanderv, line 134: For Builders, a creator called `aSomething` plays nice with the fluent api that follows. `aMessageMetadata.withIpAddress...`

osgp/protocol-adapter-iec60870/osgp-protocol-adapter-iec60870/src/main/java/org/opensmartgridplatform/adapter/protocol/iec60870/infra/messaging/DeviceRequestMessageListener.java
```diff
@@ -96,15 +95,11 @@ private void sendNotSupportedException(
...
-          new ProtocolResponseMessage.Builder()
-              .messageMetadata(messageMetadata)
-              .domain(messageMetadata.getDomain())
-              .domainVersion(messageMetadata.getDomainVersion())
+          ProtocolResponseMessage.newBuilder()
+              .messageMetadata(messageMetadata.builder().withScheduled(false).build())
               .result(ResponseMessageResultType.NOT_OK)
               .osgpException(osgpException)
               .dataObject(objectMessage.getObject())
-              .retryHeader(new RetryHeader())
```
sanderv, line 106: Did you remove this intentionally? (looks like it's removed in several locations)

bvdzwet: I did indeed remove this intentionally.
If there is a call to the builders `messageMetadata` method - as should be there in places I removed the call to `retryHeader(new RetryHeader())`, it calls `superBuilder.withMessageMetadata` which initializes the `retryHeader` with `new RetryHeader()`.
Overall I am not entirely happy with the way this `RetryHeader` is used in combination with the message metadata and response messages in getting the retries to work as intended, but improving upon it would require some more study and seemed out of scope.

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/infra/jms/MessageMetadata.java
bvdzwet: I am not entirely sure I know what you mean.
Is it replacing code like:
```java
final ProtocolResponseMessage protocolResponseMessage =
    ProtocolResponseMessage.newBuilder()
        .result(ResponseMessageResultType.NOT_OK)
        .build();
```
with something similar to:
```java
final Builder aResponseMessage = ProtocolResponseMessage.newBuilder();
final ProtocolResponseMessage protocolResponseMessage =
    aResponseMessage
        .withResult(ResponseMessageResultType.NOT_OK)
        .build();
```
If so, I don't think there is much of a reason to. Purely stylistic different preference.
If we would have some agreed standard, that should be followed, but we don't.
At the moment I tend to the more succinct version of the code without the explicit builder variable, and without the with-prefix in all builder methods.

If this is all based on my not understanding your comment, or you think we need to get more into this, please let me know.

sanderv: I meant renaming `newBuilder()` to `aMessageMetadata()`. Parenthesis make all the difference ;)

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/ConfigurationService.java
```diff
@@ -108,134 +111,89 @@ public ConfigurationService() {
...
     log.info(
         "setConfigurationObject for organisationIdentification: {} for deviceIdentification: {}",
-        deviceMessageMetadata.getOrganisationIdentification(),
-        deviceMessageMetadata.getDeviceIdentification());
+        messageMetadata.getOrganisationIdentification(),
+        messageMetadata.getDeviceIdentification());
 
-    final SmartMeter smartMeteringDevice =
-        this.domainHelperService.findSmartMeter(deviceMessageMetadata.getDeviceIdentification());
+    final SmartMeter smartMeter =
```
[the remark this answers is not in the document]
sanderv: We discussed this: not for now...

sanderv (approved):

osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/ConfigurationService.java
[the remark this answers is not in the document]
bvdzwet: Possibly an opportunity to refactor further. Two things stopped me: one is that at some point you stop refactoring, or you run the risk of going on too long, the other is that these are mostly one liners, where I could imagine that for instance with the builder there could reappear some variation again, and it would have to be fit in the refactored code. Could be I am mistaken in my assessment of what could happen, and the refactoring is still a good idea, but then my first argument still is there.

carlogeenen (approved):

PR #694, opened 2021-08-12
Feature/slim-1927 and slim-1928 hstn fw update of a smr5 E and G meter

harrymiddelburgetall:
This PR is the rework of existing but non-functioning update procedure for the E-meter.
A more clear usage of the attributes of FirmwareFile have been introduces:
- **(Firmware)Identification** is the unique public key to identify the FirmwareFile
- **ImageIdentifier** is the bytearray (stored in db in hex string) to initialize the transfer connection to the meter.

In the previous version _FirmwareFile_ were identified by the version of the _FirmwareModule_(s) the FW would update. In the current version thats simplified. _FirmwareFile_ is just identified by it's _Identification_
The _FirmwareFile_ is expected to reside in the GXF database before update can take place. The _FirmwareFile_ must also define the _DeviceModel_s it supports. For each _Device_ being updated the match with supported _DeviceModel_ is checked.

There are no seperate requesttypes for updating E-meters or G-meters. A generic UpdateFirmware request takes the _FirmwareFileIdentification_ and the _DeviceIdentification. In the adapter-domain-smartmetering layer a check is performed if the _Device_ in the request has a _GatewayDevice_. In case it has the request is directed to the _GatewayDevice_. If not the request is directed to the device identified by the _DeviceIdentification_ in the request.

There is an important differance between handling updates of E-meters and G-meters. First of all the _FirmwareFile_'s content. This is generic for all _Devices_ of the suppported _DeviceModel_s where as for the G-meters the _FirmwareFile_'s content has a header that needed to specify the _M-Bus device IdentificationNumber_ it is sent to. So after getting the _FirmwareFile_ from the _CachingRepository_ it has to be made specific for the one Device being updated.
The second difference is the _ImageIdentifier_ that is needed to initiate the connecting to the meter before transfering the _FirmwareFile_ to the meter. For G-meters this _ImageIdentifier_ is standard for all SMR5 M-Bus Devices and can be generated from information in the header of the FirmwareFile. For E-meter the _ImageIdentifier_ might differ between manfacturers and protocols (DSMR4, SMR5) there the _ImageIdentifier_ for E-meter is not generated but must be provided as information of the _FirmwareFile_ and reside in the GXF database before update can take place.


harrymiddelburgetall:
This PR has been replaced by PR #699

PR #711, opened 2021-09-14
feature/slim-1928-fw-update-G-meter-firmware-file-validation

harrymiddelburgetall:
Removed validation on Header Version in M-Bus firmware file

Signed-off-by: Harry Middelburg <harry.middelburg@alliander.com>


bvdzwet (changes requested):
I understand these changes are made because of actual firmware files that have a header with a header version other than `00h` as specified in the SMR 5.1 P2 documentation.
In that case removing the check looks appropriate.
I would suggest either logging something about versions other than specified, or at least leaving a comment in the code, so that whoever works on this code at a later time, will see that the check on the header version is not overlooked, but was intentionally removed.

bvdzwet (approved):

PR #717, opened 2021-09-16
feature/slim-1928-fw-update-G-meter-firmware-mbus-identification-number-fix

harrymiddelburgetall:
There is an important difference in the specification of the M-Bus Client Setup (Class ID: 72, OBIS-code: 0-x:24.1.0.255) between DSMR4 and SMR5
In DSMR4.2 the M-Bus Client Setup has version 0 and the identification_number attribute is defined as double-long unsigned **interpreted as BCD** where in SMR5 the M-Bus Client Setup has version 1 and the identification_number attribute is simply defined as double-long unsigned.

The value of identification_number attribute in the database is according to this difference in version

The **IdentificationNumber** handles the conversion of this database value depending on the version of the DlmsObject defined in the **DlmsObjectConfig**

### DIFF
--- a/osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/firmware/UpdateFirmwareCommandExecutor.java
+++ b/osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/firmware/UpdateFirmwareCommandExecutor.java
@@ -153,6 +153,9 @@ public class UpdateFirmwareCommandExecutor
 
     final FirmwareFile firmwareFile = new FirmwareFile(firmwareFileByteArray);
     if (firmwareFile.isMbusFirmware()) {
+
+      firmwareFile.checkLengths();
+
       this.addMac(
           messageMetadata, updateFirmwareRequestDto.getDeviceIdentification(), firmwareFile);
     }
--- a/osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/firmware/firmwarefile/FirmwareFile.java
+++ b/osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/firmware/firmwarefile/FirmwareFile.java
@@ -55,25 +55,26 @@ public class FirmwareFile {
         && this.getHeader().getAddressTypeEnum() == AddressType.MBUS_ADDRESS;
   }
 
-  public void checkLengths() {
+  public void checkLengths() throws ProtocolAdapterException {
     final FirmwareFileHeader header = this.getHeader();
     final Integer firmwareImageLength = header.getFirmwareImageLengthInt();
     final Integer securityLength = header.getSecurityLengthInt();
     final Integer headerLength = header.getHeaderLengthInt();
     if (this.imageData.length != (firmwareImageLength + securityLength + headerLength)) {
-      log.warn(
-          "Byte array length doesn't match lengths defined in header: "
-              + "\nByte array length : {}"
-              + "\nLengths defined in header: "
-              + "\nHeader : {}"
-              + "\nFirmwareImage : {}"
-              + "\nSecurity : {}"
-              + "\nTotal of {}  bytes.",
-          this.imageData.length,
-          headerLength,
-          firmwareImageLength,
-          securityLength,
-          (firmwareImageLength + securityLength + headerLength));
+      throw new ProtocolAdapterException(
+          String.format(
+              "Byte array length doesn't match lengths defined in header: "
+                  + "\nByte array length : %d"
+                  + "\nLengths defined in header: "
+                  + "\nHeader : %d"
+                  + "\nFirmwareImage : %d"
+                  + "\nSecurity : %d"
+                  + "\nTotal of %d  bytes.",
+              this.imageData.length,
+              headerLength,
+              firmwareImageLength,
+              securityLength,
+              (firmwareImageLength + securityLength + headerLength)));
     }
   }
 


osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/firmware/UpdateFirmwareCommandExecutor.java
```diff
@@ -158,30 +157,30 @@ private FirmwareFile addMac(
...
           String.format(
               EXCEPTION_MSG_DEVICE_HAS_NO_MBUS_IDENTIFICATION_NUMBER, deviceIdentification));
     }
 
-    final Long mbusIdentificationNumber =
-        new IdentificationNumber(String.format("%08d", hexNotationOfTheMbusIdentificationNumber))
-            .getIdentificationNumber();
+    // The MbusIdentificationNumber from the osgp_adapter_protocol_dlms database
+    // is used directly as int value in FirmwareFile header and subsequently the ImageIdentifier.
+    // This only works for SMR5+ devices
```
bvdzwet, line 179: This may have to do for now, in order to be able to get some progress fixing the M-Bus firmware update.
This will have to be re-assessed, and possibly changed, if we have more clarity about how the identification number is to be treated.

Plainly stating to use the value "from the database" will need to be revisited as well (possibly not immediately).
"From the database" does not tell anything about the meaning/treatment in DLMS/DSMR/SMR and is subject to potential change that should not affect the functionality here unnoticed. The way the data are stored in the DB should be an implementation detail.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/firmware/enums/FirmwareTransferStatus.java
```diff
@@ -0,0 +1,41 @@
+package org.opensmartgridplatform.adapter.protocol.dlms.domain.commands.firmware.enums;
```
bvdzwet, line 1: This source file is missing a license header.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/firmware/enums/FirmwareTransferStatus.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+public enum FirmwareTransferStatus {
```
bvdzwet, line 6: Having an enum for the status is a good idea. I think it would be good to add a little documentation linking this enum to the DLMS documentation.

It looks like it would be better to change the names a bit to more closely reflect how they are in the DLMS Blue Book with the descripton of attribute 6 image_transfer_status of the Image Transfer interface class.
The current names match what is described in SMR 5.1 P2 as firmware upgrade states, but I don't why the terms are different from DLMS, and the ImageTransfer class in the GXF DLMS protocol adapter is not only used with M-Bus firmware upgrades.

bvdzwet (changes requested):

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/firmware/enums/FirmwareTransferStatus.java
bvdzwet: I noticed an ImageTransferStatus enum already exists:
org.opensmartgridplatform.adapter.protocol.dlms.domain.commands.firmware.ImageTransfer.ImageTransferStatus
It looks like there is no need to add another one.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/firmware/UpdateFirmwareCommandExecutor.java
harrymiddelburgetall: There is a difference in th specification of the **_M-Bus Client Setup_** (Class ID: 72, OBIS-code: 0-x:24.1.0.255) between DSMR4 and SMR5
In **DSMR4.2** the _M-Bus Client Setup_ has **version 0** and the _identification_number_ attribute is defined as double-long unsigned **interpreted as BCD** where in **SMR5** the _M-Bus Client Setup_ has **version 1** and the _identification_number_ attribute is defined as double-long unsigned
So the value from the database is undefined. It depends on the version in the DlmsObjectConfig how to interpret this value. The conversion of the database value to the Numerical or Textual representation is handled by the IdentificationNumber object.

```
  private IdentificationNumber getIdentificationNumber(final DlmsDevice mbusDevice)
      throws ProtocolAdapterException {
    final Long mbusIdentificationNumberFromDatabase = mbusDevice.getMbusIdentificationNumber();

    final IdentificationNumber identificationNumber;

    if (this.mbusIdentificationIsBcdRepresentatyionAsLong(mbusDevice)) {

      identificationNumber =
          IdentificationNumber.fromBcdRepresentationAsLong(mbusIdentificationNumberFromDatabase);

    } else {
      identificationNumber =
          IdentificationNumber.fromNumericalRepresentation(mbusIdentificationNumberFromDatabase);
    }
    return identificationNumber;
  }

  private boolean mbusIdentificationIsBcdRepresentatyionAsLong(final DlmsDevice mbusDevice)
      throws ProtocolAdapterException {

    final DlmsObject mbusClientSetupObject =
        this.dlmsObjectConfigService.getDlmsObject(mbusDevice, DlmsObjectType.MBUS_CLIENT_SETUP);

    return mbusClientSetupObject.getVersion().equals(DlmsClassVersion.VERSION_0);
  }
```

PR #734, opened 2021-11-01
SMHE-344 Development: replace audit and osgp logging for elastic sear…

natasjaalliander:
…ch logging

add json logging option in core, protocoladapter and shared

### DIFF
--- a/osgp/platform/osgp-adapter-ws-shared/src/main/java/org/opensmartgridplatform/adapter/ws/infra/jms/LoggingMessageSender.java
+++ b/osgp/platform/osgp-adapter-ws-shared/src/main/java/org/opensmartgridplatform/adapter/ws/infra/jms/LoggingMessageSender.java
@@ -29,7 +29,7 @@ public class LoggingMessageSender {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(LoggingMessageSender.class);
 
-  @Value("${application.createjsonmessage:false}")
+  @Value("${auditlogging.message.create.json:false}")
   private boolean createJsonMessage;
 
   @Autowired
--- a/osgp/platform/osgp-adapter-ws-shared/src/main/java/org/opensmartgridplatform/adapter/ws/infra/jms/LoggingMessageSender.java
+++ b/osgp/platform/osgp-adapter-ws-shared/src/main/java/org/opensmartgridplatform/adapter/ws/infra/jms/LoggingMessageSender.java
@@ -29,7 +29,7 @@ public class LoggingMessageSender {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(LoggingMessageSender.class);
 
-  @Value("${application.createJsonMessage:false}")
+  @Value("${application.createjsonmessage:false}")
   private boolean createJsonMessage;
 
   @Autowired
--- a/osgp/platform/osgp-adapter-ws-shared/src/main/java/org/opensmartgridplatform/adapter/ws/infra/jms/LoggingMessageSender.java
+++ b/osgp/platform/osgp-adapter-ws-shared/src/main/java/org/opensmartgridplatform/adapter/ws/infra/jms/LoggingMessageSender.java
@@ -8,13 +8,6 @@
  */
 package org.opensmartgridplatform.adapter.ws.infra.jms;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
-import javax.jms.JMSException;
-import javax.jms.Message;
-import javax.jms.ObjectMessage;
-import javax.jms.Session;
-import javax.jms.TextMessage;
-import org.opensmartgridplatform.shared.infra.jms.Constants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -42,62 +35,19 @@ public class LoggingMessageSender {
    * @param loggingMessage The LoggingRequestMessage request message to send.
    */
   public void send(final LoggingRequestMessage loggingMessage) {
-    LOGGER.debug("Sending logger message");
-    this.sendMessage(loggingMessage);
-  }
+    if (loggingMessage == null) {
+      LOGGER.error("LoggingMessage is null and will not be send");
+    } else {
+      LOGGER.debug("Sending logger message");
+      final MessageCreator messageCreator;
+      if (this.isCreateJsonMessage()) {
+        messageCreator = new LoggingJsonMessageCreator(loggingMessage);
+      } else {
+        messageCreator = new LoggingObjectMessageCreator(loggingMessage);
+      }
 
-  /**
-   * Method for sending a logging message to the logger queue.
-   *
-   * @param loggingMessage The LoggingRequestMessage request message to send.
-   */
-  private void sendMessage(final LoggingRequestMessage loggingMessage) {
-    LOGGER.info("Sending logger message to queue");
-
-    this.loggingJmsTemplate.send(
-        new MessageCreator() {
-
-          @Override
-          public Message createMessage(final Session session) throws JMSException {
-            if (isCreateJsonMessage()) {
-              return getJsonMessage(loggingMessage, session);
-            }
-            return getObjectMessage(session, loggingMessage);
-          }
-        });
-  }
-
-  public TextMessage getJsonMessage(
-      final LoggingRequestMessage loggingMessage, final Session session) {
-    TextMessage textMessage = null;
-    try {
-      final ObjectMapper mapper = new ObjectMapper();
-      final String jsonString = mapper.writeValueAsString(loggingMessage);
-      textMessage = session.createTextMessage(jsonString);
-    } catch (final Exception e) {
-      LOGGER.error("Error creating json message : {}", e.getMessage());
+      this.loggingJmsTemplate.send(messageCreator);
     }
-    return textMessage;
-  }
-
-  public ObjectMessage getObjectMessage(
-      final Session session, final LoggingRequestMessage loggingMessage) throws JMSException {
-    final ObjectMessage objectMessage = session.createObjectMessage();
-    objectMessage.setJMSCorrelationID(loggingMessage.getCorrelationUid());
-    objectMessage.setLongProperty(Constants.TIME_STAMP, loggingMessage.getTimeStamp().getTime());
-    objectMessage.setStringProperty(Constants.CLASS_NAME, loggingMessage.getClassName());
-    objectMessage.setStringProperty(Constants.METHOD_NAME, loggingMessage.getMethodName());
-    objectMessage.setStringProperty(
-        Constants.ORGANISATION_IDENTIFICATION, loggingMessage.getOrganisationIdentification());
-    objectMessage.setStringProperty(Constants.USER_NAME, loggingMessage.getUserName());
-    objectMessage.setStringProperty(
-        Constants.APPLICATION_NAME, loggingMessage.getApplicationName());
-    objectMessage.setStringProperty(
-        Constants.DEVICE_IDENTIFICATION, loggingMessage.getDeviceIdentification());
-    objectMessage.setStringProperty(Constants.RESPONSE_RESULT, loggingMessage.getResponseResult());
-    objectMessage.setIntProperty(
-        Constants.RESPONSE_DATA_SIZE, loggingMessage.getResponseDataSize());
-    return objectMessage;
   }
 
   private boolean isCreateJsonMessage() {
--- a/osgp/platform/osgp-adapter-ws-shared/src/main/java/org/opensmartgridplatform/adapter/ws/infra/jms/LoggingMessageSender.java
+++ b/osgp/platform/osgp-adapter-ws-shared/src/main/java/org/opensmartgridplatform/adapter/ws/infra/jms/LoggingMessageSender.java
@@ -37,17 +37,17 @@ public class LoggingMessageSender {
   public void send(final LoggingRequestMessage loggingMessage) {
     if (loggingMessage == null) {
       LOGGER.error("LoggingMessage is null and will not be send");
-    } else {
-      LOGGER.debug("Sending logger message");
-      final MessageCreator messageCreator;
-      if (this.isCreateJsonMessage()) {
-        messageCreator = new LoggingJsonMessageCreator(loggingMessage);
-      } else {
-        messageCreator = new LoggingObjectMessageCreator(loggingMessage);
-      }
-
-      this.loggingJmsTemplate.send(messageCreator);
+      return;
     }
+    LOGGER.debug("Sending logger message");
+    final MessageCreator messageCreator;
+    if (this.isCreateJsonMessage()) {
+      messageCreator = new LoggingJsonMessageCreator(loggingMessage);
+    } else {
+      messageCreator = new LoggingObjectMessageCreator(loggingMessage);
+    }
+
+    this.loggingJmsTemplate.send(messageCreator);
   }
 
   private boolean isCreateJsonMessage() {
--- a/osgp/platform/osgp-adapter-ws-shared/src/main/java/org/opensmartgridplatform/adapter/ws/infra/jms/LoggingMessageSender.java
+++ b/osgp/platform/osgp-adapter-ws-shared/src/main/java/org/opensmartgridplatform/adapter/ws/infra/jms/LoggingMessageSender.java
@@ -59,10 +59,10 @@ public class LoggingMessageSender {
 
           @Override
           public Message createMessage(final Session session) throws JMSException {
-            if (LoggingMessageSender.this.isCreateJsonMessage()) {
-              return LoggingMessageSender.this.getJsonMessage(loggingMessage, session);
+            if (isCreateJsonMessage()) {
+              return getJsonMessage(loggingMessage, session);
             }
-            return LoggingMessageSender.this.getObjectMessage(session, loggingMessage);
+            return getObjectMessage(session, loggingMessage);
           }
         });
   }
--- a/osgp/platform/osgp-adapter-ws-shared/src/main/java/org/opensmartgridplatform/adapter/ws/infra/jms/LoggingMessageSender.java
+++ b/osgp/platform/osgp-adapter-ws-shared/src/main/java/org/opensmartgridplatform/adapter/ws/infra/jms/LoggingMessageSender.java
@@ -67,7 +67,7 @@ public class LoggingMessageSender {
         });
   }
 
-  private TextMessage getJsonMessage(
+  public TextMessage getJsonMessage(
       final LoggingRequestMessage loggingMessage, final Session session) {
     TextMessage textMessage = null;
     try {
@@ -80,7 +80,7 @@ public class LoggingMessageSender {
     return textMessage;
   }
 
-  private ObjectMessage getObjectMessage(
+  public ObjectMessage getObjectMessage(
       final Session session, final LoggingRequestMessage loggingMessage) throws JMSException {
     final ObjectMessage objectMessage = session.createObjectMessage();
     objectMessage.setJMSCorrelationID(loggingMessage.getCorrelationUid());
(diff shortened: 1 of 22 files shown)


osgp/platform/osgp-adapter-ws-shared/src/main/java/org/opensmartgridplatform/adapter/ws/infra/jms/LoggingMessageSender.java
```diff
@@ -50,28 +56,51 @@ private void sendMessage(final LoggingRequestMessage loggingMessage) {
...
-                Constants.DEVICE_IDENTIFICATION, loggingMessage.getDeviceIdentification());
-            objectMessage.setStringProperty(
-                Constants.RESPONSE_RESULT, loggingMessage.getResponseResult());
-            objectMessage.setIntProperty(
-                Constants.RESPONSE_DATA_SIZE, loggingMessage.getResponseDataSize());
-            return objectMessage;
+            if (LoggingMessageSender.this.isCreateJsonMessage()) {
+              return LoggingMessageSender.this.getJsonMessage(loggingMessage, session);
+            }
+            return LoggingMessageSender.this.getObjectMessage(session, loggingMessage);
```
kroesctrl, line 65: LoggingMessageSender. can be removed 3 times

kroesctrl (changes requested):
Please add tests for the changes in the following classes:
LoggingMessageSender.java
DlmsLogItemRequestMessageSender.java

osgp/platform/osgp-adapter-ws-shared/src/main/java/org/opensmartgridplatform/adapter/ws/infra/jms/LoggingMessageSender.java
natasjaalliander: really weird, this comes somehow from the formatting rules on save.

natasjaalliander:
tests are added

osgp/platform/osgp-adapter-ws-shared/src/main/java/org/opensmartgridplatform/adapter/ws/infra/jms/LoggingMessageSender.java
```diff
@@ -26,6 +29,9 @@
 
   private static final Logger LOGGER = LoggerFactory.getLogger(LoggingMessageSender.class);
 
+  @Value("${application.createJsonMessage:false}")
```
jboon, line 32: Application properties are not camelCase and it would be nice to reflect some meaning/context in the property name so it is obvious that it controls the (audit) logging messages.

natasjaalliander: application properties are lowercase now. It is actually just a setting that results in creating json messages rather than serialized messages. It can be used for other purposes than just the audit logging as well so it seems to me the naming reflects the functionality of it quite well.

kroesctrl (approved):

bvdzwet (changes requested):
Besides from the comments with the code I'd like to address an aspect that is not in the code.

These changes leave existing functionality intact, as long as configuration is not updated, and different runtime set ups are used when `auditlogging.message.create.json` is `false` (the osgp-logging component can be deployed and will deal with object messages on the logging queue) or when `auditlogging.message.create.json` is `true` (the osgp-logging component should not be deployed and the logging queue is to be listened on by something like an elastic search logging component).

This could mean that if you ever change configuration to perform JSON text message logging or not messages could be on a queue that could not be processed and might lead to runtime errors (or possibly end up in a dead letter queue because of these errors after some retries - I have not checked what would actually happen).

To prevent such issues, the listeners on the logging queue could be made aware that it may either contain `ObjectMessage` instances or `TextMessage` instances (as subtype of the `Message`).
A full implementation of both versions may probably not be worth the effort if this will not be used, but it might be good to go into `onMessage(Message message)` of `LoggingMessageListener` and `ProtocolLogItemRequestMessageListener` (assuming these are affected by the changes to the senders - should be obvious from osgp-logging.log when deployed with JSON logging) and log a warning if the message is a `TextMessage` mentioning text messages belong with a configuration setting (`auditlogging.message.create.json=true`) that is not supported by osgp-logging, and perhaps the JSON text from the message.

natasjaalliander:
Barts remarks from monday (2021-11-08) are processed: 
- Separate creators for object and json messages have been created.
- jsmtype is added to json messages
- type jsm is fixed
- I agree the tests could be improved and extended. My suggestion would be to create a separate story to improve and extend unit tests coverage for the creation and sending of messages, since this would take some effort.

natasjaalliander:
retest this please

osgp/platform/osgp-core/src/main/java/org/opensmartgridplatform/core/infra/messaging/CoreLogItemJsonMessageCreator.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright 2017 Smart Society Services B.V.
```
bvdzwet, line 2: This copyright header is not the correct one for new code.

osgp/platform/osgp-core/src/main/java/org/opensmartgridplatform/core/infra/messaging/CoreLogItemJsonMessageCreator.java
```diff
@@ -0,0 +1,34 @@
...
+package org.opensmartgridplatform.core.infra.messaging;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import javax.jms.Session;
+import javax.jms.TextMessage;
+import org.opensmartgridplatform.shared.infra.jms.Constants;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CoreLogItemJsonMessageCreator {
```
bvdzwet, line 18: Why apply different styles of message creator classes for similar use cases with message senders?
The implementation choices with the `LoggingJsonMessageCreator` where the class named something ending with `MessageCreator` actually implements the `MessageCreator` interface looks cleaner than the choices made here.

osgp/platform/osgp-core/src/main/java/org/opensmartgridplatform/core/infra/messaging/CoreLogItemJsonMessageCreator.java
```diff
@@ -0,0 +1,34 @@
...
+  public TextMessage getJsonMessage(
+      final CoreLogItemRequestMessage coreLogItemRequestMessage, final Session session) {
+    TextMessage textMessage = null;
+    try {
+      final ObjectMapper mapper = new ObjectMapper();
+      final String jsonString = mapper.writeValueAsString(coreLogItemRequestMessage);
+      textMessage = session.createTextMessage(jsonString);
+      textMessage.setJMSType(Constants.CORE_LOG_ITEM_REQUEST);
+    } catch (final Exception e) {
+      LOGGER.error("Error creating json message : {}", e.getMessage());
```
bvdzwet, line 30: Failing fast instead of only logging the exception and continuing the code flow with a most likely unusable `null` response seems more appropriate.

osgp/platform/osgp-core/src/main/java/org/opensmartgridplatform/core/infra/messaging/CoreLogItemObjectMessageCreator.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright 2017 Smart Society Services B.V.
```
bvdzwet, line 2: This copyright header is not the correct one for new code.

osgp/platform/osgp-core/src/main/java/org/opensmartgridplatform/core/infra/messaging/CoreLogItemObjectMessageCreator.java
```diff
@@ -0,0 +1,39 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.opensmartgridplatform.core.infra.messaging;
+
+import javax.jms.JMSException;
+import javax.jms.ObjectMessage;
+import javax.jms.Session;
+import org.opensmartgridplatform.shared.infra.jms.Constants;
+
+public class CoreLogItemObjectMessageCreator {
```
bvdzwet, line 16: An implementation more similar to `LoggingObjectMessageCreator` implementing `MessageCreator` looks more consistent and appropriate.

osgp/platform/osgp-core/src/main/java/org/opensmartgridplatform/core/infra/messaging/CoreLogItemRequestMessageSender.java
```diff
@@ -8,42 +8,42 @@
...
 
+  @Value("${auditlogging.message.create.json:false}")
+  private boolean createJsonMessage;
+
   @Autowired private JmsTemplate coreLogItemRequestsJmsTemplate;
 
   public void send(final CoreLogItemRequestMessage coreLogItemRequestMessage) {
+    if (coreLogItemRequestMessage == null) {
+      LOGGER.error("coreLogItemRequestMessage is null and will not be send");
+    } else {
```
bvdzwet, line 29: Better ad a return statement, separating argument checking from other logic as suggested with more detailed explanation in comments with `LoggingMessageSender`.

osgp/platform/osgp-core/src/main/java/org/opensmartgridplatform/core/infra/messaging/CoreLogItemRequestMessageSender.java
```diff
@@ -8,42 +8,42 @@
...
+      LOGGER.debug("Sending CoreLogItemRequestMessage");
 
-    LOGGER.debug("Sending CoreLogItemRequestMessage");
+      this.coreLogItemRequestsJmsTemplate.send(
```
bvdzwet, line 32: Better change this to be more like the implementation in `LoggingMessageSender` which looks a lot cleaner.

osgp/platform/osgp-core/src/test/java/org/opensmartgridplatform/core/infra/messaging/CoreLogItemJsonMessageCreatorTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+  @Test
+  void createJsonMessage() throws JMSException {
+    final TextMessage expectedTextMessage = mock(TextMessage.class);
+    when(this.session.createTextMessage(any())).thenReturn(expectedTextMessage);
+
+    final Message actualMessage =
+        this.jsonMessageCreator.getJsonMessage(this.coreLogItemRequestMessage, this.session);
+
+    assertThat(actualMessage).isSameAs(expectedTextMessage);
```
bvdzwet, line 49: Add verification that the JMSType was set on the message (see comments with `LoggingJsonMessageCreatorTest` for more details.

osgp/platform/osgp-core/src/test/java/org/opensmartgridplatform/core/infra/messaging/CoreLogItemJsonMessageCreatorTest.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  @Test
+  void jsonMessageIsEmptyWhenInputMessageHasNoValues() {
+    final Message actualMessage =
+        this.jsonMessageCreator.getJsonMessage(
+            new CoreLogItemRequestMessage(null, null, null), this.session);
+
+    assertThat(actualMessage).isNull();
+  }
```
bvdzwet, line 66: Returning `null` as a `Message` when the input is missing entirely or is missing required values to be useful passes the problem. Failing fast is very likely a better way to deal with these scenarios.

osgp/platform/osgp-core/src/test/java/org/opensmartgridplatform/core/infra/messaging/CoreLogItemObjectMessageCreatorTest.java
```diff
@@ -0,0 +1,55 @@
...
+  @Test
+  void createObjectMessage() throws JMSException {
+    final ObjectMessage expectedObjectMessage = mock(ObjectMessage.class);
+    when(this.session.createObjectMessage()).thenReturn(expectedObjectMessage);
+
+    final Message actualMessage =
+        this.objectMessageCreator.getObjectMessage(this.coreLogItemRequestMessage, this.session);
+
+    assertThat(actualMessage).isSameAs(expectedObjectMessage);
+  }
```
bvdzwet, line 49: No verification of the JMSType is performed (see similar comments in other message creator test classes for more details).

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/infra/messaging/DlmsLogItemRequestJsonMessageCreator.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public TextMessage getJsonMessage(final Session session) {
+    TextMessage textMessage = null;
+    try {
+      final ObjectMapper mapper = new ObjectMapper();
+      final String jsonString = mapper.writeValueAsString(this.dlmsLogItemRequestMessage);
+      textMessage = session.createTextMessage(jsonString);
+      textMessage.setJMSType(Constants.DLMS_LOG_ITEM_REQUEST);
+    } catch (final Exception e) {
+      LOGGER.error("Error creating json message : {}", e.getMessage());
```
bvdzwet, line 47: Logging the exception and passing `null` as a result here poses other code with the problem of getting unusable input, failing fast and have other code deal with an exception immediately looks more appropriate.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/infra/messaging/DlmsLogItemRequestMessageSender.java
```diff
@@ -31,46 +27,27 @@
...
+  @Value("${auditlogging.message.create.json:false}")
+  private boolean createJsonMessage;
 
-    this.jmsTemplate.send(new DlmsLogItemRequestMessageCreator(dlmsLogItemRequestMessage));
-  }
-
-  private static final class DlmsLogItemRequestMessageCreator implements MessageCreator {
+  public void send(final DlmsLogItemRequestMessage dlmsLogItemRequestMessage) {
+    if (dlmsLogItemRequestMessage == null) {
+      LOGGER.error("dlmsLogItemRequestMessage is null and will not be send");
```
bvdzwet, line 35: Separate the argument assertion from the logic dealing with proper input (see comments in other message sender classes for more detailed reasoning).

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/infra/messaging/DlmsLogItemRequestJsonMessageCreatorTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+  @Test
+  void jsonMessageIsEmptyWhenExceptionIsThrown() throws JMSException {
+    when(this.session.createTextMessage(any())).thenThrow(new JMSException("test jms exception"));
+
+    final Message actualMessage =
+        this.dlmsLogItemRequestJsonMessageCreator.getJsonMessage(this.session);
+
+    assertThat(actualMessage).isNull();
+  }
```
bvdzwet, line 70: See comments in other message creator test classes regarding test methods like these.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/infra/messaging/DlmsLogItemRequestObjectMessageCreatorTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Test
+  void createObjectMessage() throws JMSException {
+    final ObjectMessage expectedObjectMessage = mock(ObjectMessage.class);
+    when(this.session.createObjectMessage()).thenReturn(expectedObjectMessage);
+
+    final Message actualMessage =
+        this.dlmsLogItemRequestObjectMessageCreator.getObjectMessage(this.session);
+
+    assertThat(actualMessage).isSameAs(expectedObjectMessage);
```
bvdzwet, line 49: JMSType setting is not verified (see comments in other message creator test classes for more detailed guidance).

bvdzwet (changes requested):
Besides the comments made with the changes in this pull request, the issue of existing logging queue listeners that could get into trouble when TextMessages are in the queue instead of ObjectMessages is not addressed.
I think simply making sure an error with some explanation is logged in case of TextMessages instead of having class cast exceptions would be in order. Has a test been done with JSON formatted logging and deployment of the osgp-logging component to at least assess the impact?

natasjaalliander:
- setup is similar now for core/dlms and log message creators
- jsmtype is checked in tests
- about returning an empty message in getJsonMessage: I have separted the exceptions that can be thrown. So now a JMSException is thrown which could already be thrown by the createMessage function.
  In an earlier phase, empty messages are already checked and now only the JsonPrcessingException is catched. I don't think it is likely this will happen very often, but only then it will return an empty message. I agree passing empty messages is not nice, but the createMessage only accepts JMSExceptions.

bvdzwet (changes requested):
There is one new issue with the `CoreLogItemRequestMessageSender` that only ever creates JSON messages, no matter the configuration.

Concerns expressed in the first and second review about what happens if osgp-logging is deployed with JSON format for logging configured are neither handled nor discussed.

If you agree that an exception is appropriate in cases where a non-runtime-exception (like `JsonProcessingException` in the message creators) is thrown while the context only allows for one other specific checked exception (`JMSException` in the message creators), you have more options than logging the exception and returning a non-desirable dubious result.
As the `JsonProcessingException` extends `IOException` you could use `UncheckedIOException` as a wrapper, with something like:
```java
    } catch (final JsonProcessingException e) {
      throw new UncheckedIOException("Error processing log item as JSON text", e);
    }
```

osgp/platform/osgp-core/src/main/java/org/opensmartgridplatform/core/infra/messaging/CoreLogItemRequestMessageSender.java
```diff
@@ -8,42 +8,39 @@
...
+    if (coreLogItemRequestMessage == null) {
+      LOGGER.error("CoreLogItemRequestMessage is null and will not be send");
+      return;
+    }
     LOGGER.debug("Sending CoreLogItemRequestMessage");
+    final MessageCreator messageCreator;
+    if (this.isCreateJsonMessage()) {
+      messageCreator = new CoreLogItemJsonMessageCreator(coreLogItemRequestMessage);
+    } else {
+      messageCreator = new CoreLogItemJsonMessageCreator(coreLogItemRequestMessage);
```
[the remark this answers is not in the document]
natasjaalliander: Foutje,.. this is fixed now.

natasjaalliander:
- The Concerns about what happens if osgp-logging is deployed with JSON format for logging configured are discussed and a check is added to the listeners.
- An UncheckedIOException is thrown now instead of returning an empty message.

osgp/platform/osgp-core/src/main/java/org/opensmartgridplatform/core/infra/messaging/CoreLogItemRequestMessageSender.java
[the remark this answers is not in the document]
bvdzwet: The fix looks good.
Please keep in mind the language of communication here is English, and try to avoid mixing in Dutch words.

bvdzwet (approved):

PR #739, opened 2021-11-09
Fixed 2 connection issues in the protocol adapter.

kroesctrl:
If the invocationcounter is out of sync the DlmsConnectionManager tries to close connections that are not opened yet, this is caused by: the change in DlmsConnectionFactory where the DlmsConnectionManager is now used with try-with-resources in the method 
createAndHandleConnectionWithSecurityLevel that closes the connections.
Second issue was that when the invocationcounter was not in sync the value was not saved to the database. This is caused by a change in the DeviceRequestMessageProcessor-> processMessageTasks(). in the finally block is fresh device retrieved from the database. Before this changes the updated device with updated invocationCounter value was used, and therefor saved
```    } finally {
      final DlmsDevice device = this.domainHelperService.findDlmsDevice(messageMetadata);
      this.doConnectionPostProcessing(device, connectionManager, messageMetadata);
    }
```

### DIFF
--- a/osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/services/DomainHelperService.java
+++ b/osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/services/DomainHelperService.java
@@ -72,7 +72,12 @@ public class DomainHelperService {
   }
 
   public DlmsDevice findDlmsDevice(final MessageMetadata messageMetadata) throws OsgpException {
-    final String deviceIdentification = messageMetadata.getDeviceIdentification();
+    return this.findDlmsDevice(
+        messageMetadata.getDeviceIdentification(), messageMetadata.getIpAddress());
+  }
+
+  public DlmsDevice findDlmsDevice(final String deviceIdentification, final String ipAddress)
+      throws OsgpException {
     final DlmsDevice dlmsDevice =
         this.dlmsDeviceRepository.findByDeviceIdentification(deviceIdentification);
     if (dlmsDevice == null) {
@@ -83,14 +88,14 @@ public class DomainHelperService {
           FunctionalExceptionType.UNKNOWN_DEVICE, ComponentType.PROTOCOL_DLMS);
     }
 
-    final String ipAddress;
-    if (dlmsDevice.isIpAddressIsStatic()) {
-      ipAddress = messageMetadata.getIpAddress();
-    } else {
-      ipAddress = this.getDeviceIpAddressFromSessionProvider(dlmsDevice);
-    }
     /* Ip Address is a transient attribute for DlmsDevice, so save is not required */
-    dlmsDevice.setIpAddress(ipAddress);
+    if (dlmsDevice.isIpAddressIsStatic()) {
+      dlmsDevice.setIpAddress(ipAddress);
+    } else {
+      final String ipAddressFromSessionProvider =
+          this.getDeviceIpAddressFromSessionProvider(dlmsDevice);
+      dlmsDevice.setIpAddress(ipAddressFromSessionProvider);
+    }
     return dlmsDevice;
   }
 
--- a/osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/services/DomainHelperService.java
+++ b/osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/services/DomainHelperService.java
@@ -89,8 +89,8 @@ public class DomainHelperService {
     } else {
       ipAddress = this.getDeviceIpAddressFromSessionProvider(dlmsDevice);
     }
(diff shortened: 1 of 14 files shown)


osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/factories/DlmsConnectionManagerTest.java
```diff
@@ -0,0 +1,85 @@
...
+    assertThat(this.manager.getConnection()).isEqualTo(dlmsConnection);
+
+    this.manager.close();
+
+    assertThat(this.manager.isConnected()).isFalse();
+  }
+
+  @Test
+  void closeWithoutConnectionShouldNotFail() throws OsgpException {
+    this.manager.close();
```
carlogeenen, line 67: Please verify the absence of a connection by:
this.manager.isConnected(); 
before calling .close()

bvdzwet (commented):
If I remember correctly, having the deviceRepository in the InvocationCounterManager to save the device in the past lead to exception stack traces in the logs.
If someone can confirm the code works on an actual device with an out of sync invocation counter, without avoidable exceptions, the changes look OK (at least for now).

I am a bit worried about existing code in `org.opensmartgridplatform.adapter.protocol.dlms.infra.messaging.DeviceRequestMessageProcessor.processMessageTasks(Serializable, MessageMetadata, DlmsConnectionManager)` (lines 149..152):
```java
    // ...
    } finally {
      final DlmsDevice device = this.domainHelperService.findDlmsDevice(messageMetadata);
      this.doConnectionPostProcessing(device, connectionManager, messageMetadata);
    }
    // ...
```
If this was only done because the device variable is not in scope in the finally block, this had better been solved by moving the declaration and null-initialization of the device to before opening of the try block.
The method `findDlmsDevice(MessageMetadata)` has side effects for devices that do not have a static IP address, The session provider may be asked for an IP address for a session with the device by its IccId. This is something that helps at the start of setting a connection, but I think it may harm when called before closing the connection.
If removal of this findDlmsDeviceMethod from the finally block works as well, I think it may be a nicer solution than putting the deviceRepository in the InvocationCounterManager. If the latter works without exceptions in the logs I think it is an OK enough solution though.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/services/DomainHelperService.java
```diff
@@ -88,11 +83,14 @@ public DlmsDevice findDlmsDevice(final String deviceIdentification, final String
...
+    final String ipAddress;
     if (dlmsDevice.isIpAddressIsStatic()) {
-      dlmsDevice.setIpAddress(ipAddress);
+      ipAddress = messageMetadata.getIpAddress();
     } else {
-      dlmsDevice.setIpAddress(this.getDeviceIpAddressFromSessionProvider(dlmsDevice));
+      ipAddress = this.getDeviceIpAddressFromSessionProvider(dlmsDevice);
     }
+    dlmsDevice.setIpAddress(ipAddress);
+    this.dlmsDeviceRepository.save(dlmsDevice);
```
kroesctrl, line 93: I would like to move the update of the ip-address to a separate method.........

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/factories/DlmsConnectionHelper.java
```diff
@@ -149,6 +149,7 @@ private void createAndHandleConnectionForDevice(
             messageMetadata,
             permit,
             taskForConnectionManager);
+        return;
```
kroesctrl, line 152: Without this return always a exeception is thrown when a invocation counter is out-of-sync

bvdzwet (changes requested):
Most of the code looks like good to necessary improvements.
If my worries about the change in the `RecoverKeyProcess` regarding whether the message metadata always holds the IP address that is also in the instance field are taken away I have no further doubts.

bvdzwet:
retest this please

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/threads/RecoverKeyProcess.java
```diff
@@ -122,7 +122,7 @@ public void run() {
 
   private DlmsDevice findDevice() {
     try {
-      return this.domainHelperService.findDlmsDevice(this.deviceIdentification, this.ipAddress);
+      return this.domainHelperService.findDlmsDevice(this.messageMetadata);
```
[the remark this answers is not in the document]
kroesctrl: I totally agree, i revert the changes on the RecoverKeyProcess and added the two argument method

bvdzwet (approved):

PR #742, opened 2021-11-18
Lazy initialization of Throttling client, because it requires a http …

kroesctrl:
Lazy initialization of Throttling client, because it requires a http connection to the throttling application


osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/application/config/ThrottlingClientConfigTest.java
```diff
@@ -0,0 +1,82 @@
...
+  /*
+   * This bean should be annotated with @Lazy, because on initialization
+   * the client will register itself at the throttling service.
+   * Registering is done by a http call, whenever the throttling service is
+   * not available yet, the application should start anyway.
+   */
+  @Test
+  void testShouldHaveLazyAnnotation() throws NoSuchMethodException {
+    final Method throttlingClientMethod =
+        ThrottlingClientConfig.class.getDeclaredMethod("throttlingClient", null);
```
bvdzwet, line 51: The `null` argument seems a bit odd. It apparently has the same result as leaving it out, but `getDeclaredMethod("throttlingClient")` looks like a cleaner way for this method call without vararg arguments than `getDeclaredMethod("throttlingClient", null)`.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/config/ThrottlingClientConfig.java
```diff
@@ -10,15 +10,22 @@
...
 import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.web.client.ResourceAccessException;
 
+@Slf4j
 @Configuration
 public class ThrottlingClientConfig {
 
+  private static final int REGISTER_RETRY_DELAY = 5000;
```
bvdzwet, line 27: Why not something like:
```java
  @Value("#{T(java.time.Duration).parse('${throttling.register.retry.delay:PT5S}')}")
  private Duration registerRetryDelay;
```
It does not need to be added to the properties per se, since the default will do, at least for now, but it enables overriding without code changes, if some other value may turn out to be more desirable.
It would look more consistent with all things being configurable in this class, but it may be better if you changed this, to rename the current `delay` (perhaps something like `permitRejectedDelay`) to set the two delays apart.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/config/ThrottlingClientConfig.java
```diff
@@ -54,10 +71,27 @@ public ThrottlingClient throttlingClient() {
...
+
+    this.registerThrottlingClient(throttlingClient);
     return throttlingClient;
   }
 
+  protected void registerThrottlingClient(final ThrottlingClient throttlingClient) {
+    try {
+      log.info("Try to register the client to the throttling service");
+      throttlingClient.register();
+    } catch (final ResourceAccessException resourceAccessException) {
```
bvdzwet, line 83: Wouldn't it be better to wrap this for any kind of exception? Something similar goes for whether any exception should ever be thrown from this method (like the `BeanInitializationException` below.

The reason I comment about this, is that it is not clear to me what the consequence of any exception being thrown from the lazy bean method `throttlingClient` would be for the chances of the protocol adapter to be able to use throttling (and with that function properly).
If the bean in such case never will be initialized, the protocol adapter should no longer be running.
If the bean will be initialized on a later call (not sure how the Spring proxying handles this) there is another question of whether the code calling the bean method handles the exception in an appropriate manner.
Such code never had to deal with issues in `throttlingClientConfig.throttlingClient()` with an autowired `throttlingClientConfig`, because the application would not have started if this caused errors.

bvdzwet (changes requested):

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/application/config/ThrottlingClientConfigTest.java
kroesctrl: agree, removed null argument

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/config/ThrottlingClientConfig.java
kroesctrl: Changed both properties and added it to properties file. I prefer all configurable properties to be in the default properties file

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/config/ThrottlingClientConfig.java
kroesctrl: Agree, i think this exception should not been thrown at all, refactored it

bvdzwet: As `ResourceAccessException` is not the only exception that can possibly be thrown from `throttlingClient.register()`, and given the earlier reasoning, we don't want any exception escaping here, would it not be better to catch the more generic `Exception` here?

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/config/ThrottlingClientConfig.java
```diff
@@ -54,16 +71,33 @@ public ThrottlingClient throttlingClient() {
...
+      log.info("Throttling service was not reachable, retry in {} ms", this.registerRetryDelay);
+      try {
+        Thread.sleep(this.registerRetryDelay.toMillis());
+        this.registerThrottlingClient(throttlingClient);
```
bvdzwet, line 87: Inside this catch block, the intention is to sleep and always do a recursive invocation of `this.registerThrottlingClient(throttlingClient);`. Also in case when sleep throws `InterruptedException`.
But with this code registerThrottlingClient runs to an end in case of an interrupt, leaving an unregistered client.

One way to solve that would be to move the recursive invocation into en finally block with this try.

Looking at this code now there is something else that does not appear to be very nice.
When you want to make sure you stay out of trouble using recursion you should recurse with some changes that guarantee the recursion won't continue forever (or until stack overflow, whichever comes first).

While the current situation will probably take a long time to overflow the stack, a cleaner way to handle this repetition would be some kind of while-loop instead of recursion.
To achieve this you could make `registerThrottlingClient` return a boolean value, true if registration was called without exceptions, and false in all other cases. From `throttlingClient` you could call this in a while loop until it returns true. That would avoid flooding the stack.

bvdzwet (changes requested):

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/config/ThrottlingClientConfig.java
kroesctrl: We could do that, but i personally prefer to be as specific as possible.

bvdzwet: I know that it usually is a good idea to use that preference as a guideline, but in this specific case I understand we do not want **any** exception to pop up from the bean creation, and if so, we had better make sure that no exception escapes by catching them all. (As far as I see it, we do not have any better way to deal with any specific exception sub types, other than to keep trying to register.)

What could happen if this only catches `ResourceAccessException`? If `throttlingClient.register` throws `RestClientException` for instance, the bean method throws an exception.
What would be the result of that? That the protocol adapter remains unusable until somehow it is detected that it does not process any requests? If you go down the road of listing all specific exceptions that are currently thrown form any path in `throttlingClient.register` then you would set yourself up for trouble at some later time with code changes outside of the scope of this `ThrottlingClientConfig`.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/config/ThrottlingClientConfig.java
```diff
@@ -48,22 +50,19 @@ public String configurationName() {
...
-    final ThrottlingClient throttlingClient =
-        new ThrottlingClient(
-            new org.opensmartgridplatform.throttling.api.ThrottlingConfig(
-                this.configurationName, this.configurationMaxConcurrency),
-            this.throttlingServiceUrl,
-            this.timeout);
-    throttlingClient.register();
-    return throttlingClient;
+    return new ThrottlingClient(
+        new org.opensmartgridplatform.throttling.api.ThrottlingConfig(
```
bvdzwet, line 54: I just noticed the fully qualified reference to the `ThrottlingConfig` constructor is probably a leftover from earlier versions of the code were `ThrottlingClientConfig` was called `ThrottlingConfig` as well.
If you do something about the other comment you could consider removing the package name and importing the `ThrottlingConfig` type.

osgp/shared/osgp-throttling-client/src/main/java/org/opensmartgridplatform/throttling/ThrottlingClient.java
```diff
@@ -108,7 +128,14 @@ private void registerThrottlingConfig() {
     LOGGER.info("Registered {}", this.throttlingConfig);
   }
 
-  private void registerClient() {
+  /*
+   * registerThrottlingConfig should be synchronized, to make sure there is only
```
bvdzwet, line 132: Copy paste issue: registerThrottlingConfig should be registerClient here.

osgp/shared/osgp-throttling-client/src/main/java/org/opensmartgridplatform/throttling/ThrottlingClient.java
```diff
@@ -121,6 +148,11 @@ private void registerClient() {
 
   /** Lets the Throttling REST service know this client is going away. */
   public void unregister() {
+    if (this.clientId == null) {
+      LOGGER.info("Client was not registered yet, so skip unregistration");
```
bvdzwet, line 152: The comment certainly applies to the most common case when unregistration is skipped.
Because the `clientId` can also be `null` if `unregister` is called multiple times, it might be more accurate, to only mention that the `ThrottlingClient` does not have a registered `clientId` without the yet.

osgp/shared/osgp-throttling-client/src/main/java/org/opensmartgridplatform/throttling/ThrottlingClient.java
```diff
@@ -248,6 +280,11 @@ public Permit requestPermitUsingNetworkSegmentIfIdsAreAvailable(
   }
 
   private Integer numberOfGrantedPermits(final int requestId) {
+    if (!this.register()) {
+      LOGGER.error(
+          "Client could not be registered before requesting permit using requestId {}", requestId);
```
bvdzwet, line 285: In line with the other comment about yet, it might be better to change "Client could not be registered before" into something like "Client is not registered when" in the messages if `register` returns `false`.
It may not be likely with the current way the `ThrottlingClient` is used, but if some later change would trigger this scenario the remark that registration has not happened could become misleading.

osgp/shared/osgp-throttling-client/src/test/java/org/opensmartgridplatform/throttling/ThrottlingClientTest.java
```diff
@@ -61,16 +61,8 @@ private MockResponse okWithIdResponse(final int id) {
         .setBody(String.valueOf(id));
   }
 
-  @Test
-  void throttlingConfigAndClientAreRegisteredWithTheThrottlingService() {
-    final short throttlingConfigId = 83;
-    final int clientId = 42;
-    this.whenTheThrottlingServiceReturnsIdsOnRegistration(throttlingConfigId, clientId);
```
bvdzwet, line 68: It looks like this was the only place from which private method `whenTheThrottlingServiceReturnsIdsOnRegistration` was called. If it is no longer of use to the test it had better been removed as well.

osgp/shared/osgp-throttling-client/src/test/java/org/opensmartgridplatform/throttling/ThrottlingClientTest.java
```diff
@@ -168,6 +185,47 @@ void clientRequestsPermitByNetworkSegment() {
...
+  void registerFailureClientRequestsPermitByNetworkSegment() {
+    final short throttlingConfigId = 37;
+    final int clientId = 347198;
+    final int baseTransceiverStationId = 983745;
+    final int cellId = 2;
+    final int requestId = 894;
+    this.whenTheThrottlingServiceReturnsFailureOnRegistration();
+
+    final Permit expectedPermit =
+        new Permit(throttlingConfigId, clientId, requestId, baseTransceiverStationId, cellId, null);
```
bvdzwet, line 221: The expected permit is not used and had better be removed, as the test verifies that the requested permit is not present.
The `throttlingConfigId`, and `clientId` don't belong in this test as these are only returned if the throttling service is up and functioning, which it is not in this scenario.
If you want to keep the `requestId` you had better add a call to `whenTheThrottlingClientUsesNextRequestId`, but it may be better to leave both out as the actual `requestId` appears to have no particular relevance to this scenario.

bvdzwet (changes requested):

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/config/ThrottlingClientConfig.java
```diff
@@ -10,12 +10,14 @@
...
 
 import java.time.Duration;
+import lombok.extern.slf4j.Slf4j;
 import org.opensmartgridplatform.throttling.ThrottlingClient;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
 
+@Slf4j
```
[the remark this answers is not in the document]
kroesctrl: Agree, removed annotation

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/config/ThrottlingClientConfig.java
kroesctrl: changed

osgp/shared/osgp-throttling-client/src/main/java/org/opensmartgridplatform/throttling/ThrottlingClient.java
kroesctrl: changed

osgp/shared/osgp-throttling-client/src/main/java/org/opensmartgridplatform/throttling/ThrottlingClient.java
kroesctrl: Changed

osgp/shared/osgp-throttling-client/src/main/java/org/opensmartgridplatform/throttling/ThrottlingClient.java
kroesctrl: Changed 3 times

osgp/shared/osgp-throttling-client/src/test/java/org/opensmartgridplatform/throttling/ThrottlingClientTest.java
kroesctrl: removed method

osgp/shared/osgp-throttling-client/src/test/java/org/opensmartgridplatform/throttling/ThrottlingClientTest.java
kroesctrl: Applied changes

osgp/shared/osgp-throttling-client/src/main/java/org/opensmartgridplatform/throttling/ThrottlingClient.java
bvdzwet: Why only with the `numberOfGrantedPermits` and one of the `releasePermit` methods, and not with the other `releasePermit` and `discardPermit`?

kroesctrl: Changed

bvdzwet (approved):

PR #776, opened 2022-01-17
smhe 228 prevent simultaneous replace key operations - Bundled request support

harrymiddelburgetall:
Check on running key changing processes is moved from MessageProcessor to CommandExecutor.
If process is running a DeviceKeyProcessAlreadyRunningException is thrown.
If this exception is caught in the DeviceRequestMessageProcessor the request is put back on the queue and NOT sent back the Core (and therefor NOT triggering retry by the RetryHeader).
Also removed the entanglement of _Replace_ and _GenerateAndReplace_ commandExecutors and the _execute_ and _executeBundleAction_ flow


osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/security/ReplaceKeyCommandExecutor.java
```diff
@@ -66,146 +49,29 @@ public ReplaceKeyCommandExecutor() {
...
-      final MessageMetadata messageMetadata)
-      throws OsgpException {
-
-    if (!keyWrapper.isGenerated()) {
-      this.secretManagementService.storeNewKey(
-          messageMetadata,
-          device.getDeviceIdentification(),
-          keyWrapper.getSecurityKeyType(),
-          keyWrapper.getBytes());
+      this.deviceKeyProcessingService.startProcessing(messageMetadata.getDeviceIdentification());
```
jboon, line 62: I see this execute method with the startProcessing call also in the abstract super class. Either implement it in the super and use that or remove the implementation in the abstract super and use the specific implementation.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/exceptions/DeviceKeyProcessAlreadyRunningException.java
```diff
@@ -10,13 +10,13 @@
...
-  private static final long serialVersionUID = 1998438538193678335L;
+public class DeviceKeyProcessAlreadyRunningException extends DirectlyRetryableException {
 
-  public DeviceKeyProcessAlreadyRunningException() {}
+  private static final long serialVersionUID = 1998438538193678335L;
 
-  public DeviceKeyProcessAlreadyRunningException(final String message) {
-    super(message);
+  public DeviceKeyProcessAlreadyRunningException(final Duration delay) {
+    super(delay);
```
jboon, line 20: It is uncommon to pass steering data in a exception. I would advice to use the retry mechanism that is in place already. It also contains a field to control the delay.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/infra/messaging/DeviceRequestMessageProcessor.java
```diff
@@ -113,7 +114,12 @@ public void processMessage(final ObjectMessage message) throws JMSException {
           messageObject, messageMetadata, this.throttlingClientConfig.permitRejectedDelay());
 
     } catch (final Exception exception) {
-      this.sendErrorResponse(messageMetadata, exception, messageObject);
+      if (exception instanceof DirectlyRetryableException) {
```
jboon, line 117: This should be achieved with an extra catch block. But I would not send the message directly to the queue. Please use the mechanism that is in place.

jboon (changes requested):
Use the retry mechanism and let the message return to core first and be rescheduled.
Use the logic in the abstract super of your command executor or the logic in your specific implementation, not both.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/infra/messaging/DeviceRequestMessageProcessor.java
harrymiddelburgetall: Discussion on retry mechanism in core for 'direct' retry is pending. This case of 'direct' retry is therefor solved in DLMS-adapter component (in similar way retry for throttling is handled). Possibly refactoring is needed on conclusion of pending discussion.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/exceptions/DeviceKeyProcessAlreadyRunningException.java
harrymiddelburgetall: By catching DeviceKeyProcessAlreadyRunningException in DeviceRequestMessageProcessor we have a less generic implementation but we do not need to store steering data like duration in our exception

jboon (approved):

PR #778, opened 2022-01-18
FDP-526 Use MQTT broker by configuration instead of as device

bvdzwet:
The set up with the MQTT broker as device was grown historically when at
times each known device had an internal MQTT broker.
As we currently have an MQTT broker of which lots of devices are
measurement publishing clients, the set up got confusing.
With these changes the single broker is configured in properties and not
seen as a device that is monitored and managed.
Data from actual devices published on the broker is routed to the
distribution automation Kafka adapter to be published.


sanderv:
retest this please

sanderv (approved):

PR #779, opened 2022-01-19
SMHE-565 Update datatype for mbus identification number

natasjaalliander:
revert leftpadding with zeros of mbus identification number and update test data, mbus identification number with length 7 is not a valid case


jboon (approved):

natasjaalliander:
retest this please

PR #782, opened 2022-01-26
SMHE-329 Add devices to core database

harrymiddelburgetall:
Signed-off-by: Harry Middelburg <harry.middelburg@alliander.com>


harrymiddelburgetall:
Testdata should not be inserted here

PR #796, opened 2022-02-18
SMHE-462 DEV: ENCS Headend frame counter can be lowered

natasjaalliander:
check if invocation counter is lower than current value, if so generate event

### DIFF
--- a/integration-tests/cucumber-tests-platform-smartmetering/src/test/resources/features/osgp-adapter-ws-smartmetering/monitoring/SystemEvent.feature
+++ b/integration-tests/cucumber-tests-platform-smartmetering/src/test/resources/features/osgp-adapter-ws-smartmetering/monitoring/SystemEvent.feature
@@ -12,7 +12,7 @@ Feature: SmartMetering Monitoring - System Events
       | CommunicationMethod      | GPRS              |
       | Protocol                 | SMR               |
       | ProtocolVersion          |               5.1 |
-      | InvocationCounter        |        2000000000 |
+      | InvocationCounter        |          20000000 |
       | Port                     |              1027 |
     When the get "INTERVAL" meter reads request is received
       | DeviceIdentification | TEST1027000000001        |
--- a/integration-tests/cucumber-tests-platform-smartmetering/src/test/resources/features/osgp-adapter-ws-smartmetering/monitoring/SystemEvent.feature
+++ b/integration-tests/cucumber-tests-platform-smartmetering/src/test/resources/features/osgp-adapter-ws-smartmetering/monitoring/SystemEvent.feature
@@ -4,6 +4,28 @@ Feature: SmartMetering Monitoring - System Events
   I want to be able to receive system events
   So that I am aware of and can possibly act upon these events
 
+  Scenario: Receive a system event when the invocation counter with a device reaches a threshold value
+    Given a dlms device
+      | DeviceIdentification     | TEST1029000000001 |
+      | DeviceType               | SMART_METER_E     |
+      | Hls5active               | true              |
+      | CommunicationMethod      | GPRS              |
+      | Protocol                 | SMR               |
+      | ProtocolVersion          |               5.2 |
+      | InvocationCounter        |        2000000000 |
+      | Port                     |              1029 |
+    When the get "INTERVAL" meter reads request is received
+      | DeviceIdentification | TEST1029000000001        |
+      | PeriodType           | INTERVAL                 |
+      | BeginDate            | 2015-09-01T00:00:00.000Z |
+      | EndDate              | 2015-09-01T01:00:00.000Z |
+    Then a system event should be returned
+      | DeviceIdentification | TEST1029000000001                     |
+      | SystemEventType      | INVOCATION_COUNTER_THRESHOLD_REACHED  |
+    And the "INTERVAL" meter reads result should be returned
+      | DeviceIdentification | TEST1029000000001 |
+
+
   Scenario: Receive a system event when the invocation counter with a device is lowered
     Given a dlms device
       | DeviceIdentification     | TEST1027000000001 |
--- a/integration-tests/cucumber-tests-platform-smartmetering/src/test/resources/features/osgp-adapter-ws-smartmetering/monitoring/SystemEvent.feature
+++ b/integration-tests/cucumber-tests-platform-smartmetering/src/test/resources/features/osgp-adapter-ws-smartmetering/monitoring/SystemEvent.feature
@@ -25,6 +25,25 @@ Feature: SmartMetering Monitoring - System Events
     And the "INTERVAL" meter reads result should be returned
       | DeviceIdentification | TEST1027000000001 |
 
+  Scenario: Receive a system event when the invocation counter with a device is lowered
+    Given a dlms device
+      | DeviceIdentification     | TEST1027000000001 |
+      | DeviceType               | SMART_METER_E     |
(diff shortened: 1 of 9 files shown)


harrymiddelburgetall (commented):
To me it is counter intuitive  that when you have prevented the invocation counter to be lowered you generate a system event _INVOCATION_COUNTER_LOWERED_. I expect an event like _INVOCATION_COUNTER__**NOT**__LOWERED_ or _ATTEMPT_TO LOWER_INVOCATION_COUNTER_

natasjaalliander:
INVOCATION_COUNTER_LOWERED is updated to  ATTEMPT_TO_LOWER_INVOCATION_COUNTER, indeed a much better name

harrymiddelburgetall (approved):

natasjaalliander:
retest this please

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/infra/messaging/DlmsConnectionMessageProcessor.java
```diff
@@ -147,8 +147,12 @@ protected void doConnectionPostProcessing(
     }
 
     if (device.needsInvocationCounter()) {
-      this.updateInvocationCounterForDevice(device, conn);
-
+      final boolean invocationCounterLowered =
+          this.systemEventService.receivedInvocationCounterIsLowerThanCurrentValue(
+              device, messageMetadata, conn);
```
bvdzwet, line 152: This looks like the wrong place to perform this check.

One reason would be that it does not belong as a check with all connections an invocation counter is involved, but only in those connections where GXF needs to update the invocation counter before getting a working connection (that is, where GXF can be manipulated to reuse invocation counter values).

Another issue is related to the implementation of the check, where device communication is used to retrieve the invocation counter. Besides my objections to this practice, as expressed in the `SystemEventService` review comments, there is an issue that the throttling constraints are ignored. You should not communicate with a device after releasing the throttling permit, as is done a couple of lines above this change.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/application/services/SystemEventServiceTest.java
```diff
@@ -50,6 +61,8 @@ void setUp() {
             this.osgpRequestMessageSender,
             this.correlationIdProviderService,
             this.invocationCounterEventThreshold);
+    this.device = this.getDevice();
```
bvdzwet, line 64: I don't mind extracting builder invocations into methods, but it does not feel nice when there is general set up, with a general name like `getDevice` that is actually about a specific special case (a device with its invocation counter at the threshold value).

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/application/services/SystemEventServiceTest.java
```diff
@@ -122,31 +183,77 @@ void verifyMaxValueReachedEvent() {
...
-    assertThat(metadata.getDomainVersion()).isEqualTo(messageMetadata.getDomainVersion());
+    assertThat(metadata.getDomain()).isEqualTo(this.messageMetadata.getDomain());
+    assertThat(metadata.getDomainVersion()).isEqualTo(this.messageMetadata.getDomainVersion());
+  }
+
+  @Test
+  void verifyHigherInvocationCounterReceived() throws IOException {
+    final DlmsConnection dlmsConnection = mock(DlmsConnection.class);
+    final DlmsConnectionManager connectionManager = mock(DlmsConnectionManager.class);
+    final long receivedHigherThanActualInvocationCounter = 100;
```
bvdzwet, line 220: The test gives the correct result, but there are some aspects that don't help me understand it and follow the logic.
One is the use of an invocation counter higher than the one that is used as the threshold related to another system event.
The other is that in tests where it is about the value of the invocation counter with the device entity versus the value of the invocation counter received from a physical device, it helps to see both of these values explicitly in the test method.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/infra/messaging/DlmsConnectionMessageProcessor.java
bvdzwet: In the code there are flows from
`org.opensmartgridplatform.adapter.protocol.dlms.domain.factories.DlmsConnectionHelper.createAndHandleConnectionForDevice(DlmsDevice, DlmsMessageListener, ConnectionProperties, MessageMetadata, Permit, Consumer<DlmsConnectionManager>)`
where the invocation counter is out of sync, which is then handled by:
`org.opensmartgridplatform.adapter.protocol.dlms.domain.factories.InvocationCounterManager.initializeWithInvocationCounterStoredOnDeviceTask(DlmsDevice, DlmsConnectionManager)`.
Here you'll  find:
```java
      final Long previousKnownInvocationCounter = device.getInvocationCounter();
      final Long invocationCounterFromDevice = this.getInvocationCounter(connectionManager);
```
What I would expected to see was that in this method a check would be performed if `invocationCounterFromDevice` is a lower value than `previousKnownInvocationCounter`. If so, the value should not be set on the device, in order to prevent it from being used (with the risk of reusing an invocation counter).
Probably some kind of exception should be thrown here to terminate the attempt to create the device connection.

Then, at some appropriate place in the flow the system event could be sent, and the request being handled could get an error response.
Maybe this should not even be a system event, as these are usually used to signal something separate from the web service request-response flow. In this case I would think a functional exception code could be more appropriate, where the application could translate this in an event code for other systems.

integration-tests/cucumber-tests-platform-smartmetering/src/test/resources/features/osgp-adapter-ws-smartmetering/monitoring/SystemEvent.feature
```diff
@@ -21,9 +43,7 @@ Feature: SmartMetering Monitoring - System Events
       | EndDate              | 2015-09-01T01:00:00.000Z |
     Then a system event should be returned
       | DeviceIdentification | TEST1027000000001                     |
-      | SystemEventType      | INVOCATION_COUNTER_THRESHOLD_REACHED  |
-    And the "INTERVAL" meter reads result should be returned
-      | DeviceIdentification | TEST1027000000001 |
+      | SystemEventType      | ATTEMPT_TO_LOWER_INVOCATION_COUNTER   |
```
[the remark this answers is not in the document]
bvdzwet: As one of the requirements is that the the invocation counter must not be lowered with a device if a value that is received through the public client is lower than the one stored, I think a verification step is in order, that checks that the invocation counter stored with the device at the end of the scenario is no less than the value from the set up (here 20000000).

bvdzwet (changes requested):

PR #798, opened 2022-02-23
FDP-527: Added Prometheus support

sanderv:


osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/metrics/DefaultMetricsService.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.opensmartgridplatform.shared.config.PrometheusDisabledCondition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.context.annotation.Conditional;
+import org.springframework.stereotype.Component;
+
+@Component
+@Conditional(PrometheusDisabledCondition.class)
+public class DefaultMetricsService implements MetricsService {
```
bvdzwet, line 21: Perhaps some other name would better describe what this does. Default may suggest some kind of metrics related working, but this ignores the metering part.
Another thing that might be a little confusing is that this `DefaultMetricsService` is not related to `enableDefaultMetrics` in `MetricsConfig`.

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/metrics/MetricsService.java
```diff
@@ -0,0 +1,49 @@
...
+   */
+  void increaseCounter(String name);
+
+  /**
+   * Increase the counter identified by {@code name} by 1.
+   *
+   * @param name Name of the counter
+   * @param description Description of the counter for humans
+   */
+  void increaseCounter(String name, String description);
```
bvdzwet, line 29: It feels a bit smelly that in this method it appears that on any invocation to influence the metric, a `description` is provided. I would expect the description to be related to the metric itself, and not to any specific increment of the counter. (What would it even mean if you were to call this for the same name, with different descriptions?)

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/metrics/PrometheusMetricsServer.java
```diff
@@ -67,7 +70,7 @@
   @Value("${metrics.prometheus.path}")
   private String path;
 
-  @Value("${metrics.componentname}")
+  @Value("${metrics.componentname:}")
```
bvdzwet, line 73: Looks a bit strange. If the intention is to make the component name optional (not sure if that is a good choice by itself) it looks odd to call `requireNonNull(this.componentName, "...")` in the `@PostConstruct` method `start`.
Or is there some particular reason that a blank String is OK, but `null` is not?

osgp/shared/shared/src/test/java/org/opensmartgridplatform/shared/config/PrometheusDisabledConditionTest.java
```diff
@@ -0,0 +1,57 @@
...
+  void shouldBeTrueOnMissingProperty() {
+    final MockEnvironment mockEnvironment = new MockEnvironment();
+    when(this.conditionContext.getEnvironment()).thenReturn(mockEnvironment);
+
+    assertThat(this.prometheusDisabledCondition.matches(this.conditionContext, null)).isTrue();
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"false", "no", "yes", "banana", ""})
+  void shoudBeTrueOnPropertyOtherThanTrueProperty(final String propertyValue) {
```
bvdzwet, line 41: Typo: shoud -> should (in some other method names as well).

bvdzwet (changes requested):

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/metrics/DefaultMetricsService.java
sanderv: Decided to not follow this path, class is no longer part of the solution

osgp/shared/shared/src/test/java/org/opensmartgridplatform/shared/config/PrometheusDisabledConditionTest.java
sanderv: Fixed

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/metrics/PrometheusMetricsServer.java
bvdzwet: My objection was not to making the component name optional. If this makes sense now that is OK with me.
What bothers me is, that if the component name has become optional, the following check does not seem to make sense:
```java
requireNonNull(this.componentName, "Component name not set");
```
It would appear to be more logical to leave that requirement not to be null out, and have null values treated the same as an empty String in: `if (!StringUtils.isEmpty(this.componentName)) {`.
Or is there something I am missing that would make it logical to get an error on null, demanding the component name to be set, while it has become an optional thing?

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/config/MetricsConfig.java
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Provides a Prometheus metrics server with optional default metrics for CPU, memory, database. See
+ * {@link PrometheusMetricsServer} for details on how to configure the port/path of the metrics
+ * server.
+ *
+ * <p>The provided {@link PrometheusMeterRegistry} bean can be used to add custom metrics.
+ */
+@Configuration
+// Starts the PrometheusMetricsServer
```
bvdzwet, line 44: I guess this comment had better been removed or altered. I don't understand what it clarifies here, and to which code it applies.
If it is about the component scan making sure the `PrometheusMetricsServer` component is included in the Spring context whenever the `MetricsConfig` is imported, that could be a useful comment, but starting the metrics server is dependent on code in the `PrometheusMetricsServer` and the `PrometheusEnabledCondition`.

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/config/MetricsConfig.java
```diff
@@ -0,0 +1,102 @@
...
+  private DataSource dataSource;
+
+  /**
+   * MeterRegistry to publish custom metrics. This bean is required because custom metrics depend on
+   * it.
+   *
+   * @return registry
+   */
+  @Bean
+  public PrometheusMeterRegistry getMeterRegistry() {
```
bvdzwet, line 65: It is rather unconventional for Spring bean methods to follow the Java bean getter naming convention.
Wouldn't `meterRegistry()` be a more common way to name this?

bvdzwet (changes requested):

bvdzwet (approved):

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/config/MetricsConfig.java
sanderv: Clarified this in person; it's for the @ComponentScan below. Changed the comment to be more clear

PR #802, opened 2022-03-03
Feature/smhe 462 headend framecounter lowered throw functionalexception

natasjaalliander:


osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/factories/InvocationCounterManagerTest.java
```diff
@@ -93,4 +101,41 @@ void initializesInvocationCounterForDevice() throws Exception {
...
+    invocationManagerLogger.addAppender(listAppender);
+
+    final FunctionalException thrownException =
+        catchThrowableOfType(
+            () ->
+                this.manager.initializeWithInvocationCounterStoredOnDeviceTask(
+                    this.device, connectionManager),
+            FunctionalException.class);
+
+    assertThat(logsList.size() > 0).isTrue();
```
bvdzwet, line 129: Something like: `assertThat(logsList).isNotEmpty();` would give nicer exception messages.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/factories/InvocationCounterManagerTest.java
```diff
@@ -93,4 +101,41 @@ void initializesInvocationCounterForDevice() throws Exception {
...
+    assertThat(
+            logsList
+                .get(0)
+                .toString()
+                .contains("[ERROR] Attempt to lower invocationCounter of device"))
+        .isTrue();
```
bvdzwet, line 135: As stated, I am not a fan of assertions on logging, but if you do so, I think the following would improve things a bit:
```java
    final ILoggingEvent loggingEvent = logsList.get(0);
    assertThat(loggingEvent).asString().contains("Attempt to lower invocationCounter of device");
    assertThat(loggingEvent.getLevel()).isEqualTo(Level.ERROR);
```
What I believe to be nicer in this alternative compared to the current code is that it takes out the "[ERROR]" part, and replaces it by an explicit assert on the level. The formatting of the word "ERROR" with the square brackets could change when using different logging configuration, and you would probably not want that to break the test. By using `contains` on the String assertion, you would get a nicer error message than with `isTrue` (boolean assertion) on the result of `String.contains`.

bvdzwet (changes requested):

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/factories/InvocationCounterManagerTest.java
```diff
@@ -93,4 +101,41 @@ void initializesInvocationCounterForDevice() throws Exception {
     assertThat(this.device.getVersion()).isGreaterThan(this.initialDeviceVersion);
     assertThat(this.device.getInvocationCounter()).isEqualTo(invocationCounterValueOnDevice);
   }
+
+  @Test
+  void attemptToLowerInvocationCounterForDeviceLogsErrorAndThrowsException() throws Exception {
+    final Logger invocationManagerLogger =
+        (Logger) LoggerFactory.getLogger(InvocationCounterManager.class);
+    final ListAppender<ILoggingEvent> listAppender = new ListAppender<>();
+    final List<ILoggingEvent> logsList = listAppender.list;
```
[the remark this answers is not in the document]
natasjaalliander: noted and agreed.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/factories/InvocationCounterManagerTest.java
```diff
@@ -93,4 +101,41 @@ void initializesInvocationCounterForDevice() throws Exception {
...
+    final long invocationCounterValueOnDevice = 1;
```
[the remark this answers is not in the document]
natasjaalliander: I'm not sure what you mean here but I do see that it can be made more clearer. I have tried to improve it by giving them better names, so it is more self explainable.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/factories/InvocationCounterManagerTest.java
natasjaalliander: right, I have updated this line to the proposed assertThat(logsList).isNotEmpty();

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/factories/InvocationCounterManagerTest.java
natasjaalliander: yes, should have done this in the first place. It is updated now.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/factories/InvocationCounterManagerTest.java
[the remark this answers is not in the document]
bvdzwet: The new name provides information about the invocation counter on the device.
This could help clarifying the expectation should the test fail.
With something like the following, the relevant values for the invocation counter are visible and controlled local to this single test:
```
    final long invocationCounterValueOnDevice = 1;
    final long previouslyKnownInvocationCounter = 7; // bigger than invocationCounterValueOnDevice
    this.device.setInvocationCounter(previouslyKnownInvocationCounter);
```
An alternative could be a builder method for a device where the invocation counter is a parameter, and use that to obtain the device.

Another advantage would be that somebody who makes changes to the general before-each set up of the device would not have to take into account which values are needed by which tests.
If for some other test a value of 83 would come in more convenient, this test might still succeed, but `deviceWithPreviouslyKnownInvocationCounterSeven` would no longer makes sense. This too can be avoided by taking the expected value (7) into the test implementation instead of referring to it inside the general `setUp` by naming the device variable.

bvdzwet (approved):

PR #810, opened 2022-03-14
SMHE-632 Set the correct MessageType in SetPushSms and SetPushAlarm

stefanermens:


harrymiddelburgetall (approved):
Clean separation of conditional setting of attributes (in the CommandExecutor) and the mapping to the DataObject (in the Mapper)

PR #811, opened 2022-03-15
SMHE-575 Firmware update with multiple adapters

stefanermens:


osgp/platform/osgp-core/src/main/java/org/opensmartgridplatform/core/infra/jms/protocol/outbound/ProtocolResponseMessageSender.java
```diff
@@ -32,21 +34,42 @@ public void send(
       final ProtocolInfo protocolInfo,
       final MessageMetadata messageMetadata) {
 
+    this.sendWithDestination(responseMessage, messageType, protocolInfo, messageMetadata, null);
```
natasjaalliander, line 37: Maybe use the jmsTemplate.getDefaultDestination() here, instead of using null as destination. Now you add null, and you have to deal with this null further down the road.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/infra/messaging/requests/to/core/OsgpRequestMessageSender.java
```diff
@@ -27,10 +28,29 @@
...
       final MessageMetadata messageMetadata) {
+    this.send(requestMessage, messageType, messageMetadata, false);
+  }
+
+  public void sendWithReplyToThisInstance(
+      final RequestMessage requestMessage,
+      final String messageType,
+      final MessageMetadata messageMetadata) {
+    this.send(requestMessage, messageType, messageMetadata, true);
+  }
```
natasjaalliander, line 47: Why 2 send functions? If the boolean has a good name, it should be clear, right?

kevinsmeets (approved):

osgp/platform/osgp-core/src/main/java/org/opensmartgridplatform/core/infra/jms/protocol/outbound/ProtocolResponseMessageSender.java
stefanermens: I agree, that's a better way to do it.

osgp/platform/osgp-core/src/main/java/org/opensmartgridplatform/core/infra/jms/protocol/inbound/messageprocessors/GetFirmwareFileMessageProcessor.java
```diff
@@ -83,21 +84,26 @@ public void processMessage(final ObjectMessage message) throws JMSException {
...
+      this.sendSuccessResponse(
+          metadata, device.getProtocolInfo(), firmwareFileDto, message.getJMSReplyTo());
 
     } catch (final Exception e) {
       LOGGER.error("Exception while retrieving firmware file: {}", firmwareFileIdentification);
       final OsgpException osgpException =
           new OsgpException(
               ComponentType.OSGP_CORE, "Exception while retrieving firmware file.", e);
-      this.sendFailureResponse(metadata, device.getProtocolInfo(), osgpException);
+      if (device != null) {
```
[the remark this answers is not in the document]
stefanermens: This problem already existed in the original code. Device can be null if the device can't be found. An exception will be thrown and we end up in this catch, where we try to send a failure response. For this failure response a ProtocolInfo is needed. When we don't have a device, we can't get the protocol info and we can't send the failure response.
I've added a log.error to make clear we can't send the failure response in this case.

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/infra/messaging/requests/to/core/OsgpRequestMessageSender.java
stefanermens: I try to avoid boolean parameters in public functions, because from the outside it's not immediately clear what the difference is between `send(message, true)` or `send(message, false)`. Of course, a good name for the boolean will help the reader to figure out what the boolean means, but I prefer a meaningful function name,

natasjaalliander (approved):

PR #855, opened 2022-05-25
Feature/smhe 511 protocol update

jboon:
Added smartmetering specific endpoint for Update Protocol.

Reason:
- consistent implementation of asynchronous request
- extension of existing endpoint (which is synchronous) would lead to asynchronous implementation for all users, since update of core database must take place after update of dlms database

The protocol variant field is only added to (and is relevant in) the protocol_info table in the core database. With this field we can distinguish between different protocols combined with another value (variant). In our case this column will contain the device communication method (CDMA, LTE-M, etc). With this we can configure different protocol adapter queues so we can have different protocol adapter instances depending on the communication method.


osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/infra/jms/core/messageprocessors/UpdateProtocolResponseMessageProcessor.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2017 Smart Society Services B.V.
```
kroesctrl, line 2: copyright header

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/services/ManagementService.java
```diff
@@ -142,4 +144,14 @@ public void clearMBusStatusOnAllChannels(
     this.clearMBusStatusOnAllChannelsCommandExecutor.execute(
         conn, device, clearMBusStatusOnAllChannelsRequestDto, messageMetadata);
   }
+
+  public UpdateProtocolResponseDto updateProtocol(
+      final DlmsDevice device, final UpdateProtocolRequestDto requestDto) {
+
+    device.setProtocol(requestDto.getProtocol(), requestDto.getProtocolVersion());
```
kroesctrl, line 151: Is it correct that ProtocolVariant is only used in the core, and not in the protocol adapter?

kroesctrl (changes requested):

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/application/services/ManagementService.java
jboon: Yes, that is correct. The protocol variant field is only added to and relevant in protocol_info tabel in the core database. With this field we can distinguish between different protocols combined with another value (variant). In our case this column will contain the device communication method (CDMA, LTE-M, etc). With this we can configure different protocol adapter queues so we can have different protocol adapter instances depending on the communication method.

kroesctrl (approved):

PR #918, opened 2022-10-03
FDP-842: CycloneDX SBOM plugin configured

sanderv:
- Enabled for each .war project
- bom.* is packaged in each .war for Openshift VI scanning

### DIFF
--- a/super/pom.xml
+++ b/super/pom.xml
@@ -1130,7 +1130,7 @@
     <repository>
       <id>fuse-repo</id>
       <name>Fuse Source</name>
-      <url>http://repo.fusesource.com/nexus/content/groups/public/</url>
+      <url>https://repo.fusesource.com/nexus/content/groups/public/</url>
       <releases>
         <enabled>false</enabled>
       </releases>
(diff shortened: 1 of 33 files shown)


super/pom.xml
```diff
@@ -1157,6 +1161,51 @@
...
+      <id>spring-roo-repository</id>
+      <url>https://repo1.maven.org/maven2/</url>
+      <releases>
+        <enabled>true</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+    </repository>
+-->
```
smvdheijden: Shouldn't this block be removed?

super/pom.xml
```diff
@@ -472,6 +472,10 @@
             <groupId>org.springframework</groupId>
             <artifactId>spring-aop</artifactId>
           </exclusion>
+          <exclusion>
+            <groupId>com.querydsl</groupId>
+            <artifactId>querydsl-apt</artifactId>
+          </exclusion>
```
smvdheijden: Is this exclusion needed for getting the CycloneDC SBOM plugin to work?

super/pom.xml
sanderv: removed

super/pom.xml
sanderv: removed

smvdheijden (approved):

PR #930, opened 2022-10-17
SMHE-1060 Correct filtering of meterreads when time is outside requested period

stefanermens:
This PR solves an issue with filtering meterreads when the time is outside the requested period. When the meter uses 'null' time stamps, this filtering results in removing all values instead of only the values outside the period. Example: the period requested is 00:00 - 00:00 UTC, the meter returns 22:00 - 22:00. The filtering should remove just the 2 first values. However, because the first value with the complete timestamp is filtered out, the following 'null' timestamps can't be calculated, so those values are removed as well. This is solved by moving the filtering to another place.


osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/periodicmeterreads/AbstractPeriodicMeterReadsCommandExecutor.java
```diff
@@ -232,6 +228,21 @@ private AmrProfileStatusCodeDto readAmrProfileStatusCode(final DataObject amrPro
...
+          "Not using an object from capture buffer (clock= {}), because the date does not match the given period: [ {} .. {} ].",
+          meterReadTime,
+          beginDateTime,
+          endDateTime);
+      return false;
+    } else {
+      return true;
+    }
+  }
+
```
harrymiddelburgetall, line 245: Why is this validation moved from the DlmsHelper to this AbstractCommandExecutor class?

harrymiddelburgetall (approved):

stefanermens:
Successful nigthly build: https://jenkins.fdp.osgp.cloud/view/Nightly%20Build/job/gxf-nightly-build/545

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/periodicmeterreads/AbstractPeriodicMeterReadsCommandExecutor.java
stefanermens: The validation is now on the resulting ResponseItems, so it's not related to DLMS and therefore the DlmsHelper is not the right place anymore.

PR #974, opened 2023-01-30
Move output setting to device table

jasperkamerling:
The database doesn't match the JPA model that was created. This may cause issues when deleting ssld's or devices.
The real solution would be to refactor the database but this is a first step to have the jpa model match the database.

Schema

### DIFF
--- a/osgp/platform/osgp-domain-core/src/main/java/org/opensmartgridplatform/domain/core/entities/Device.java
+++ b/osgp/platform/osgp-domain-core/src/main/java/org/opensmartgridplatform/domain/core/entities/Device.java
@@ -24,7 +24,9 @@ import java.util.TreeSet;
 import java.util.stream.Collectors;
 import javax.persistence.AttributeOverride;
 import javax.persistence.CascadeType;
+import javax.persistence.CollectionTable;
 import javax.persistence.Column;
+import javax.persistence.ElementCollection;
 import javax.persistence.Embedded;
 import javax.persistence.Entity;
 import javax.persistence.EnumType;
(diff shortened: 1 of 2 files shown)


osgp/platform/osgp-domain-core/src/main/java/org/opensmartgridplatform/domain/core/entities/Device.java
```diff
@@ -96,6 +100,15 @@ public class Device extends AbstractEntity {
...
 
+  /**
+   * The output settings are only used in the ssld table but in the database it is linked to the
+   * Device Table.
+   */
+  @LazyCollection(LazyCollectionOption.FALSE)
+  @ElementCollection()
+  @CollectionTable(name = "device_output_setting", joinColumns = @JoinColumn(name = "device_id"))
+  protected List<DeviceOutputSetting> outputSettings = new ArrayList<>();
+
```
smvdheijden, line 111: I prefer not to have an SSLD specific field in the generic device Entitiy. It would probably be better to change the relationship in the DB and change the foreign key to point to the ssld id (which should effectively be the same as the device id due to the inheritance relation between them)?

smvdheijden (changes requested):

osgp/platform/osgp-domain-core/src/main/java/org/opensmartgridplatform/domain/core/entities/Device.java
jasperkamerling: True that would be better but that would be a big change in the data model.
This just fixes the application issues with the data model.

smvdheijden: I don't think a big change in the data model is needed. Setting ON DELETE CASCADE will probably also work?

PR #977, opened 2023-01-31
Make use of Artemis jms client library possible

kroesctrl:
We had the problem: high-prio messages are not always processed quick.
The reason is that we are using PooledConnectionFactory in combination with prefetch.

https://activemq.apache.org/what-is-the-prefetch-limit-for
`
Consuming messages from a pool of consumers an be problematic due to prefetch. Unconsumed prefetched messages are only released when a consumer is closed, but with a pooled consumer the close is deferred (for reuse) till the consumer pool closes. This leaves prefetched messages unconsumed till the consumer is reused. This feature can be desirable from a performance perspective. However, it can lead to out-of-order message delivery when there is more than one consumer in the pool. For this reason, the org.apache.activemq.pool.PooledConnectionFactory does not pool consumers.
`

Solution is not using a PooledConnectionFactory when a prefetch is configured, and also we are using a Artimis jms client.
In the solution we tried to implement this with backwards compatibility. So the changes are not breaking existing implementations:

Changes required in properties to make use of this solution:
jms.default.broker.type=ARTEMIS
jms.default.broker.url=failover:(tcp://localhost:61617) -> (tcp://localhost:61617)
jms.default.connection.queue.consumer.window.size=0

Redelivery properties should be set in broker.xml

PooledConnectionFactory is not used anymore for Listeners/Consumers. 
Consumers should not use ConnectionPooling: https://activemq.apache.org/maven/apidocs/org/apache/activemq/jms/pool/PooledConnectionFactory.html

### DIFF
--- a/osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/application/config/messaging/JmsBrokerActiveMq.java
+++ b/osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/application/config/messaging/JmsBrokerActiveMq.java
@@ -47,11 +47,9 @@ public class JmsBrokerActiveMq implements JmsBroker {
   private static final Logger LOGGER = LoggerFactory.getLogger(JmsBrokerActiveMq.class);
 
   private final JmsPropertyReader propertyReader;
-  private final RedeliveryPolicy redeliveryPolicy;
 
   public JmsBrokerActiveMq(final JmsPropertyReader propertyReader) {
     this.propertyReader = propertyReader;
-    this.redeliveryPolicy = this.getRedeliveryPolicy();
   }
 
   @Override
@@ -124,7 +122,7 @@ public class JmsBrokerActiveMq implements JmsBroker {
     LOGGER.debug("Initializing redelivery policy map.");
 
     final RedeliveryPolicyMap redeliveryPolicyMap = new RedeliveryPolicyMap();
-    redeliveryPolicyMap.setDefaultEntry(this.redeliveryPolicy);
+    redeliveryPolicyMap.setDefaultEntry(this.getRedeliveryPolicy());
     return redeliveryPolicyMap;
   }
 
--- a/osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/application/config/messaging/JmsBrokerActiveMq.java
+++ b/osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/application/config/messaging/JmsBrokerActiveMq.java
@@ -1,0 +1,150 @@
+/*
+ * Copyright 2023 Alliander N.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.opensmartgridplatform.shared.application.config.messaging;
+
+import static org.opensmartgridplatform.shared.application.config.messaging.JmsPropertyNames.PROPERTY_NAME_BACK_OFF_MULTIPLIER;
+import static org.opensmartgridplatform.shared.application.config.messaging.JmsPropertyNames.PROPERTY_NAME_BROKER_CLIENT_KEY_STORE;
+import static org.opensmartgridplatform.shared.application.config.messaging.JmsPropertyNames.PROPERTY_NAME_BROKER_CLIENT_KEY_STORE_SECRET;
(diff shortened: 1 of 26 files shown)


kroesctrl:
https://jenkins.fdp.osgp.cloud/view/SMHE-cucumber/job/gxf-smhe-cucumber-build/32/

kroesctrl:
https://jenkins.fdp.osgp.cloud/view/SMHE-cucumber/job/gxf-smhe-cucumber-build/33/

abolder (approved):

kroesctrl:
https://jenkins.fdp.osgp.cloud/view/SMHE-cucumber/job/gxf-smhe-cucumber-build/34/

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/application/config/messaging/JmsBrokerArtemis.java
```diff
@@ -0,0 +1,116 @@
...
+import org.apache.activemq.artemis.jms.client.ActiveMQConnectionFactory;
+import org.apache.activemq.artemis.jms.client.ActiveMQQueue;
+import org.apache.commons.lang3.NotImplementedException;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** This class provides the basic components used for JMS messaging. */
+public class JmsBrokerArtemis implements JmsBroker {
+  private static final Logger LOGGER = LoggerFactory.getLogger(JmsBrokerActiveMq.class);
```
jasperkamerling: The logger does not match the class

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/application/config/messaging/JmsBrokerActiveMq.java
```diff
@@ -0,0 +1,150 @@
...
+public class JmsBrokerActiveMq implements JmsBroker {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(JmsBrokerActiveMq.class);
+
+  private final JmsPropertyReader propertyReader;
+  private final RedeliveryPolicy redeliveryPolicy;
+
+  public JmsBrokerActiveMq(final JmsPropertyReader propertyReader) {
+    this.propertyReader = propertyReader;
+    this.redeliveryPolicy = this.getRedeliveryPolicy();
```
jasperkamerling: I don't really think we need a field for this function. And using an overridable method in a constructor is discouraged.

jasperkamerling:
I'm also missing some unit tests on the new configuration code

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/application/config/messaging/JmsBrokerArtemis.java
kroesctrl: Changed

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/application/config/messaging/JmsBrokerActiveMq.java
kroesctrl: Agree, changed

kroesctrl (commented):
Unit tests are added

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/application/config/messaging/JmsPropertyNames.java
```diff
@@ -28,6 +29,8 @@ public final class JmsPropertyNames {
       "connection.pool.time.between.expiration.check.millis";
   public static final String PROPERTY_NAME_CONNECTION_POOL_IDLE_TIMEOUT =
       "connection.pool.idle.timeout";
+  public static String PROPERTY_NAME_CONNECTION_QUEUE_CONSUMER_WINDOW_SIZE =
```
jasperkamerling: Should be final

jasperkamerling (changes requested):

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/application/config/messaging/JmsPropertyNames.java
kroesctrl: aangepast

jasperkamerling (approved):

PR #982, opened 2023-02-09
SMHE-1286: Add encryption for LLS

stefanermens:


stefanermens:
This PR updates the LLS1 connector to be able to connect to DSMR2.2 smart meters. DSMR2.2 uses 
- LLS1 with a password 
- and encryption and authentication for each message sent or received, except for the AARQ.
 
Note that:
- Changes to the jDLMS library are needed as well. When this new version is available, some additional changes might be necessary, for example to set some option to make the library work as we want for DSMR2.2.
- The LLS1 connector is not a generic LLS1 connector. There might be a need for a more generic LLS1 connector in the future, but for now we decided to make it work like this.

kroesctrl (approved):

PR #986, opened 2023-02-23
Logging shows message, current version of flyway does not support pos…

kroesctrl:
Logging shows message, current version of flyway does not support postgres version 12, so upgrade library


kroesctrl:
https://jenkins.fdp.osgp.cloud/view/SMHE-cucumber/job/gxf-smhe-cucumber-build/40/

PR #996, opened 2023-03-17
FDP-993: Added health check mechanism

sanderv:
…entations


osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/health/DataSourceHealthCheck.java
```diff
@@ -0,0 +1,49 @@
...
+      this.dataSource = null;
+      LOGGER.info("No datasource found for HealtCheck");
+    }
+  }
+
+  @Override
+  public HealthResponse isHealthy() {
+    try {
+      if (this.dataSource != null) {
+        this.dataSource.getConnection();
```
jasperkamerling: It might be good to run the isValid check on the connection to make sure it is not closed.

kevinsmeets (approved):

PR #1009, opened 2023-04-20
Add delay for Artemis instance

kroesctrl:
The property that is required for delayed messages is different on Artemis:

ActiveMQ uses the property AMQ_SCHEDULED_DELAY is used for delay values
Artemis uses the property _AMQ_SCHED_DELIVERY is used for delay values


abolder (approved):
No remarks

PR #1010, opened 2023-04-21
SMHE-1444 DSMR2.2 on demand PQ

stefanermens:
The command executor for actual PQ (on demand PQ), requested values from too much PQ objects from DSMR2.2 meters. To solve this, a new object config is added for DSMR2.2 meters. This new config only contains the PQ objects that are present in DSMR2.2 meters. For one object the OBIS code is different in DSMR2.2.
The cucumber test is updated to test DSMR2.2 as well. The check if all expected values are received is changed from a fixed list to a list based on the number of expected objects per protocol.


harrymiddelburgetall (approved):
Ziet er goed uit.
Misschien kan je nog een commentaar regel opnemen in de feature file die uitlegt dat alleen het aantal expected quality objects in de lijst wordt gecontroleerd

PR #1024, opened 2023-06-05
chore: copyright and license annotations

nicorikken:
Improvement on copyright and license annotations using `reuse annotate`.


sanderv:
This was a test run of the licensing tool. Better results in other branches.

PR #1037, opened 2023-07-18
SMHE-1564 radius add FunctionalExceptionType

natasja-n:
SMHE-1564 : Add FunctionalExceptionType SUBSCRIPTION_INFORMATION_COULD_NOT_BE_SET.  This is used in the BundleRequestProcessor in case subscriptioninformation should be set but could not be obtained.


kroesctrl (approved):

kroesctrl (changes requested):
Can we not use ApplicationFunctionalExceptionType in SMHE, because this is a SMHE exception

natasja-n:
Close pr. Exception is added to ApplicationFunctionalExceptionType and is not needed here.

PR #1057, opened 2023-08-25
FDP-1558: Refactored DaylightSavingTimeTransition.java

LucianoFavoroso:


osgp/protocol-adapter-iec61850/osgp-protocol-adapter-iec61850/src/main/java/org/opensmartgridplatform/adapter/protocol/iec61850/domain/valueobjects/DaylightSavingTimeTransition.java
```diff
@@ -271,19 +276,24 @@ public DateTime getDateTime(
...
-        if (thirdAttempt.getMonthOfYear() > secondAttempt.getMonthOfYear()) {
+        final ZonedDateTime thirdAttempt = secondAttempt.plusDays(7);
+        if (thirdAttempt.getMonthValue() > secondAttempt.getMonthValue()) {
           return secondAttempt.plusHours(this.getTime(transition));
         }
         return thirdAttempt.plusHours(this.getTime(transition));
       }
     };
 
+    public static final int SUNDAY = 7;
```
JelleHoffman: Why here? Is would suspect that these values are at the top of the file.

JelleHoffman (changes requested):

JelleHoffman (approved):

PR #1072, opened 2023-09-18
Defineable load capture period is 900

kroesctrl:


abolder (approved):
Great that the exceptions from the specification are integrated in the profile description.

PR #1075, opened 2023-09-20
Feature/smhe 1730 poll alarm

kroesctrl:
After sending wakeup SMS to the meter by jasperwireless, we wait for the alarm that is send by the meter to connect to the meter.


abolder (approved):
No comments

PR #1092, opened 2023-10-10
Feature/smhe 1602 add scalar unit from object config

abolder:
Add a config_lookup_type column to dlms_device table for use in the new object config service. This is used for profile defined scalar units.


abolder:
https://jenkins.fdp.osgp.cloud/view/SMHE-cucumber/job/gxf-smhe-cucumber-build/135/

PR #1094, opened 2023-10-11
SMHE-1602: add config_lookup_type to dlms_device

abolder:
Add a type of meter to the dlms-device so that we can use this type to lookup specific config values from the new dlms object config service. For now it's used to type G-meters for size of connection: e.g.: G4, G6, G10, G16, G25.


abolder:
✔︎ https://jenkins.fdp.osgp.cloud/view/SMHE-cucumber/job/gxf-smhe-cucumber-build/140/

PR #1109, opened 2023-11-01
When firmware file is not available device connection is not required.

kroesctrl:
When firmware file is not available device connection is not required, because then first the firmware file needs to be retrieved form osgp-core. When a device connection is not required, there will not be send a SMS for devices without static ip-address (GPRS).

### DIFF
--- a/osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/infra/messaging/processors/UpdateFirmwareRequestMessageProcessor.java
+++ b/osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/infra/messaging/processors/UpdateFirmwareRequestMessageProcessor.java
@@ -38,6 +38,23 @@ public class UpdateFirmwareRequestMessageProcessor extends DeviceRequestMessageP
   }
 
   @Override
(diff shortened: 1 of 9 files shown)


osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/infra/messaging/processors/UpdateFirmwareRequestMessageProcessor.java
```diff
@@ -37,6 +37,23 @@ protected UpdateFirmwareRequestMessageProcessor() {
     super(MessageType.UPDATE_FIRMWARE);
   }
 
+  @Override
+  protected boolean usesDeviceConnection(final Serializable messageObject) {
+    if (messageObject instanceof final UpdateFirmwareRequestDto requestDto) {
+      final String firmwareIdentification = requestDto.getFirmwareIdentification();
+      final boolean usesDeviceConnection =
```
harrymiddelburgetall, line 44: Catching the return value of this method (isFirmwareFileAvailable) in the (usesDeviceConnection) variable hides an implicite business rule. Better make it more explicite.

harrymiddelburgetall (changes requested):

harrymiddelburgetall (approved):

PR #1118, opened 2023-11-17
Revert "SMHE-1622: Set parallel_requests_allowed to false"

stefanermens:
This reverts commit fc5fb0a4ce2c6f21b0a875acee80c9adefd8c765.


harrymiddelburgetall (commented):
If this flyway script is already merge and deployed it cannot be reverted like this I think without extra actions on the database it is deployed to.

harrymiddelburgetall (approved):

PR #1126, opened 2023-11-23
Remove local Throttling Service

kroesctrl:
Remove local Throttling Service


abolder (approved):
This is an old ThrottlingService and can indeed be removed. Because it's the probable cause for the application to hang on a deadlock.

abolder (changes requested):
This service is needed for gprs, so it would be nice if this service will be configurable by properties to be enabled of disabled.

PR #1127, opened 2023-11-27
Configurable throttling service

kroesctrl:
There can be 3 types of throttling in the protocol-adapter:
'local' -> throttling per instance of the application
'shared' -> use a the shared osgp-throttling-service, so throttling will be applied over all instances
'disabled' -> no throttling

The local variant now cleans up expired permits, after a specific time


abolder (approved):
No remarks. There are now three thottling service options available.

PR #1159, opened 2024-01-17
Bundle requests for G-meter devices without a gateway device are not …

kroesctrl:
SMHE-1767: Bundle requests for G-meter devices without a gateway device are not allowed.
Do not send requests to the protocol-adapter but fail with a FunctionalException


osgp/platform/osgp-adapter-domain-smartmetering/src/main/java/org/opensmartgridplatform/adapter/domain/smartmetering/application/services/BundleService.java
```diff
@@ -169,4 +174,17 @@ private void checkIfAdditionalActionIsNeeded(
       }
     }
   }
+
+  private void checkIfBundleRequestIsAllowed(final Device smartMeter) throws FunctionalException {
+    if (this.isGasMeter(smartMeter.getDeviceType())) {
```
jboon, line 179: DeviceType heeft een methode isEMeter()

kroesctrl, line 179: DeviceType is een String hier......

jboon (approved):

PR #1161, opened 2024-01-22
Change loglevel of message that gas model cannot be found in configur…

kroesctrl:
Change loglevel of message that gas model cannot be found in configuration

### DIFF
--- a/osgp/protocol-adapter-dlms/osgp-dlms/src/main/java/org/opensmartgridplatform/dlms/services/ObjectConfigService.java
+++ b/osgp/protocol-adapter-dlms/osgp-dlms/src/main/java/org/opensmartgridplatform/dlms/services/ObjectConfigService.java
@@ -371,10 +371,11 @@ public class ObjectConfigService {
             originalAttribute.copyWithNewValueAndType(value, ValueType.FIXED_IN_METER);
         return cosemObject.copyWithNewAttribute(newAttribute);
       } else {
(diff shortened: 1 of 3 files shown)


osgp/protocol-adapter-dlms/osgp-dlms/src/main/java/org/opensmartgridplatform/dlms/services/ObjectConfigService.java
```diff
@@ -371,10 +371,11 @@ protected CosemObject handleValueBasedOnModel(
             originalAttribute.copyWithNewValueAndType(value, ValueType.FIXED_IN_METER);
         return cosemObject.copyWithNewAttribute(newAttribute);
       } else {
-        // If no value was found for this device model, then set the value to Dynamic and log a
-        // warning. The command executor should get the value from the meter.
-        log.warn(
-            "Could not find value for devicemodel {} for {}",
+        // If no value was found for this device model, then set the value to Dynamic and log the
+        // message
+        // The command executor should get the value from the meter.
```
stefanermens, line 376: `The command executor should get the value from the meter.` This is true for the case we discussed, but could not be true in other situations.

kroesctrl: I do not understand your comment. In this case the value should be retrieved from the meter.

stefanermens (approved):

PR #1166, opened 2024-01-25
Feature/smhe 1918 tls support

kroesctrl:
add tls supported versions


loesimmens (approved):

Marcel-Jansen (approved):

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/infra/ws/DefaultWebServiceTemplateFactory.java
```diff
@@ -90,6 +91,7 @@ public static class Builder {
     private String keyStoreLocation;
     private String keyStorePassword;
     private KeyStoreFactoryBean trustStoreFactory;
+    private String[] supportedTlsProtocols = new String[] {"TLSv1.2", "TLSv1.3"};
```
harrymiddelburgetall, line 94: Shouldn't this be from configuration? Or at least be enumerated.

harrymiddelburgetall (commented):
One remark

PR #1172, opened 2024-02-15
SMHE-1841: Make get/set WithList more efficient

stefanermens:
This PR updates the withListSupported property of a device to withListMax. This new property is the maximum amount of items in a request. The getWithList function in the DlmsHelper automatically divides the list of addresses in to sublists of the maximum amount. If only 1 address is provided or if withListMax is 1, then it will use a normal DLMS GetRequest instead of GetWithList.
The database is updated. All devices with withListSupported to false are set to withListMax = 1. All devices with withListSupported to true are set to withListMax = 32, which is the default that was already used.

### DIFF
--- a/integration-tests/cucumber-tests-platform-smartmetering/src/test/resources/features/osgp-adapter-ws-smartmetering/management/SetDeviceCommunicationSettings.feature
+++ b/integration-tests/cucumber-tests-platform-smartmetering/src/test/resources/features/osgp-adapter-ws-smartmetering/management/SetDeviceCommunicationSettings.feature
@@ -13,7 +13,7 @@ Feature: SmartMetering Management - Set Device Communication Settings
       | DeviceIdentification     | TEST1024000000001 |
       | DeviceType               | SMART_METER_E     |
       | ChallengeLength          | 8                 |
-      | WithListSupported        | true              |
+      | WithListMax              | 32                |
       | SelectiveAccessSupported | true              |
       | IpAddressIsStatic        | true              |
       | UseSn                    | true              |
@@ -22,7 +22,7 @@ Feature: SmartMetering Management - Set Device Communication Settings
     When the set device communication settings request is received
       | DeviceIdentification     | TEST1024000000001 |
       | ChallengeLength          | 16                |
-      | WithListSupported        | false             |
+      | WithListMax              | 1                 |
       | SelectiveAccessSupported | false             |
       | IpAddressIsStatic        | false             |
       | UseSn                    | false             |
@@ -31,7 +31,7 @@ Feature: SmartMetering Management - Set Device Communication Settings
     Then the set device communication settings response should be "OK"
     And the device "TEST1024000000001" should be in the database with attributes
       | ChallengeLength          | 16    |
-      | WithListSupported        | false |
+      | WithListMax              | 1     |
       | SelectiveAccessSupported | false |
       | IpAddressIsStatic        | false |
       | UseSn                    | false |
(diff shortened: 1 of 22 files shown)


osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/utils/DlmsHelperTest.java
```diff
@@ -80,36 +83,99 @@ public class DlmsHelperTest {
...
-  public void testGetWithListWorkaround() throws ProtocolAdapterException, IOException {
+  @ParameterizedTest
+  @CsvSource({"0", "1"})
+  void testNormalGetWithMultipleAddresses(final int getWithListMax)
+      throws ProtocolAdapterException, IOException {
     final DlmsConnection dlmsConnection = mock(DlmsConnection.class);
     final DlmsConnectionManager connectionManager = mock(DlmsConnectionManager.class);
     final DlmsDevice dlmsDevice = mock(DlmsDevice.class);
+
+    // Set the maximum to 0 or 1. When the max is 0, then it will be handled as a max of 1.
```
harrymiddelburgetall, line 146: Is this a test for some hidden feature? or do we miss setting this max value somewhere?

integration-tests/cucumber-tests-platform-smartmetering/src/test/resources/features/osgp-adapter-ws-smartmetering/management/SetDeviceCommunicationSettings.feature
```diff
@@ -13,7 +13,7 @@ Feature: SmartMetering Management - Set Device Communication Settings
       | DeviceIdentification     | TEST1024000000001 |
       | DeviceType               | SMART_METER_E     |
       | ChallengeLength          | 8                 |
-      | WithListSupported        | true              |
+      | WithListMax              | 32                |
```
harrymiddelburgetall, line 16: Since 32 was the default value used when with list supported was true I would feel comfortable with a second test with a different max value.

harrymiddelburgetall (commented):
Nice change, two small remarks

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/test/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/utils/DlmsHelperTest.java
harrymiddelburgetall: Okay, this makes it fool proof

integration-tests/cucumber-tests-platform-smartmetering/src/test/resources/features/osgp-adapter-ws-smartmetering/management/SetDeviceCommunicationSettings.feature
harrymiddelburgetall: Okay, this is covered by unit tests

stefanermens: This test is for making sure that setting the value works. Testing here with another value is not necessary.
In the test for the DlmsHelper there are some tests with different values to verify the correct request(s) are sent.

harrymiddelburgetall (approved):

PR #1173, opened 2024-02-15
SMHE-191: An external SIG analysis fix

kroesctrl:
SMHE-191: An external SIG analysis, recommended to make this not a public class.

This class could be vulnerable to arbitrary command execution. The reason is that this class executes a command on the operating system.


jboon (approved):

osgp/shared/shared/src/main/java/org/opensmartgridplatform/shared/infra/networking/ping/Pinger.java
```diff
@@ -81,4 +81,218 @@ public boolean ping(final String destination) {
     }
     return false;
   }
+
+  /*
+   * An external SIG analysis, recommended to make this not a public class.
+   * This class could be vulnerable to arbitrary command execution.
+   * The reason is that this class executes a command on the operating system.
+   */
+  class CommandExecutor {
```
jboon, line 90: Deze zou nog private gemaakt kunnen worden.

kroesctrl: Makes the testability a lot harder, so i leave it package private

PR #1339, opened 2024-06-07
Feature/smhe 2018 invalid identification number

harrymiddelburgetall:
This PR solves a problem when decoupling mbus device when channel information cannot be correctly interpreted.
In case the channel information was not corrently interpretable then an exception was raised before clearing the channel.
Now the channel is cleared in both cases. If not correctly interpretable the mbus identification number was sent back with the response as raw meter data.


natasja-n (dismissed):

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/mbus/IdentificationNumber.java
```diff
@@ -63,12 +63,10 @@ public static IdentificationNumber fromNumericalRepresentation(
     return new IdentificationNumber(String.format("%08d", identificationInNumericalRepresentation));
   }
 
-  private static Long toBcdRepresentationAsLong(
-      final String identificationNumberInTextualRepresentation) {
+  private Long toBcdRepresentationAsLong(final String identificationNumberInTextualRepresentation) {
```
stefanermens: Why remove `static` while still getting the value from the parameter? Suggestion: remove the parameter and use `this.textualRepresentation`.

stefanermens (changes requested):

osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/mbus/IdentificationNumber.java
harrymiddelburgetall: Implemented according to suggestion

stefanermens (approved):

PR #1445, opened 2024-11-11
Redis connection pool size is now configurable

jboon:


natasja-n (dismissed):

harrymiddelburgetall (changes requested):
Should this property also be added to the properties file?

jboon:
A default value of 64 is defined in the configuration class. This is/was also the default for the Redis client (Redisson). The property will be overridden in the deployment environment with environment variables, so defining them in the property file seems unnecessary.

harrymiddelburgetall (approved):

PR #1486, opened 2024-12-13
SMHE-2412: Remove old object config

stefanermens:
This PR removes the old object config (in `osgp/protocol-adapter-dlms/osgp-protocol-adapter-dlms/src/main/java/org/opensmartgridplatform/adapter/protocol/dlms/domain/commands/dlmsobjectconfig`).
The old config was still used for the PushNotifications. This is changed to use the new object config.
And finally, some enums are still in use. They are moved to the new object config.


osgp/protocol-adapter-dlms/osgp-dlms/src/main/java/org/opensmartgridplatform/dlms/objectconfig/CommunicationMethod.java
```diff
@@ -0,0 +1,33 @@
...
+
+  public static CommunicationMethod getCommunicationMethod(final String method)
+      throws ObjectConfigException {
+    return switch (method) {
+      case "GPRS" -> GPRS;
+      case "CDMA" -> CDMA;
+      case "LTE" -> LTE;
+      default -> throw new ObjectConfigException("Unknown communication method: " + method);
+    };
+  }
```
harrymiddelburgetall: 1-on-1 copy: looks like a lot of code for enum.valueOf()

harrymiddelburgetall (changes requested):
Nice cleanup. Just one remark to cleanup even more.

harrymiddelburgetall (approved):

PR #1534, opened 2025-02-28
FDP-2849: Docker image creation changed to remove version from web-path

sanderv:
@BuildImages


sanderv:
closed, merged to other branch

PR #1546, opened 2025-03-19
FDP-2849: Updating tags docker metadata. Latest is not set correctly

JelleHoffman:
Signed-off-by: Jelle Hoffman jelle.hoffman@hotmail.com


sanderv (approved):

paulusalliander (commented):
Erg cool, om de latest tag te genereren op deze manier, via de raw type.

PR #1687, opened 2025-09-25
FDP-3268 ~ Adds protocol info for oslp mikronika

smvdheijden:


osgp/platform/osgp-core/src/main/resources/db/migration/V20250925090535000__Adds_OSLP_Mikronika_protocol_info.sql
```diff
@@ -0,0 +1,43 @@
+-- SPDX-FileCopyrightText: 2025 Contributors to the GXF project
+--
+-- SPDX-License-Identifier: Apache-2.0
+DO
+$$
+BEGIN
+
+IF NOT EXISTS (
```
jessiekrul, line 8: Is this really needed when using flyway?

osgp/platform/osgp-core/src/main/resources/db/migration/V20250925090535000__Adds_OSLP_Mikronika_protocol_info.sql
```diff
@@ -0,0 +1,43 @@
+-- SPDX-FileCopyrightText: 2025 Contributors to the GXF project
+--
+-- SPDX-License-Identifier: Apache-2.0
+DO
+$$
```
jessiekrul, line 5: I've never seen this before, is this something specific for this db?

JKruisheer (approved):

osgp/platform/osgp-core/src/main/resources/db/migration/V20250925090535000__Adds_OSLP_Mikronika_protocol_info.sql
smvdheijden: Maybe not, but it wouldn't harm in this case. We have been using it in the flyway scripts for a long time...
(This file is actually a modified copy of an existing protocol info migration)

osgp/platform/osgp-core/src/main/resources/db/migration/V20250925090535000__Adds_OSLP_Mikronika_protocol_info.sql
smvdheijden: `$$` is a PostgreSQL-specific delimiter used for writing multi-line string literals.

osgp/platform/osgp-core/src/main/resources/db/migration/V20250925090535000__Adds_OSLP_Mikronika_protocol_info.sql
jessiekrul: So because we've been using this for a long time means we can't improve?

osgp/platform/osgp-core/src/main/resources/db/migration/V20250925090535000__Adds_OSLP_Mikronika_protocol_info.sql
jessiekrul: In this case the whole query? Or what is the multi-line string literal in this case?

jessiekrul (approved):

osgp/platform/osgp-core/src/main/resources/db/migration/V20250925090535000__Adds_OSLP_Mikronika_protocol_info.sql
smvdheijden: Everything within the `$$`. It's recommended when using DO, see also: https://www.postgresql.org/docs/current/sql-do.html

osgp/platform/osgp-core/src/main/resources/db/migration/V20250925090535000__Adds_OSLP_Mikronika_protocol_info.sql
smvdheijden: I think there has been a valid reason to use this in the past. Can't exactly remember anymore what it might have been though...

PR #1701, opened 2025-10-10
SMHE-2342 Job Clear Alarm Register overschrijft Alarm Registers allee…

natasja-n:
…n wanneer status groter is dan nul


natasja-n:
commits were not signed off, this pr is replaced by an other with correct commits

PR #1750, opened 2026-01-20
Feature/fdp 3615 serialization

jessevdent:


osgp/platform/osgp-core/src/test/java/org/opensmartgridplatform/core/infra/jms/protocol/outbound/serialization/ResponseDeserializationTest.java
```diff
@@ -0,0 +1,58 @@
...
+   * Verifies that a {@link DeviceStatusDto} object serialized by the
+   * gxf-publiclighting-message-transformer project can be successfully deserialized and interpreted
+   * in this project.
+   *
+   * <p>The serialized input file device-status-response.ser is produced in the transformer project
+   * by running the CreateSerializedObjectsTest. Copy that file into this project's {@code
+   * src/test/resources/} directory.
+   */
+  @Test
+  void testDeserializationDeviceStatusTest() {
```
smvdheijden, line 23: When having so many assertions, you could consider creating an expected device status dto and using an assertion using recursive comparison.

osgp/platform/osgp-core/src/test/java/org/opensmartgridplatform/core/infra/jms/protocol/outbound/serialization/CreateSerializedObjectsTest.java
```diff
@@ -0,0 +1,120 @@
...
+    configurationDto.setAstroGateSunSetOffset(141);
+    configurationDto.setSwitchingDelays(asList(142, 143));
+    configurationDto.setRelayLinking(createRelayMatrixDto());
+    configurationDto.setRelayRefreshing(true);
+    configurationDto.setSummerTimeDetails(
+        ZonedDateTime.ofInstant(
+            Instant.ofEpochMilli(146L * 24 * 60 * 60 * 1000), ZoneId.systemDefault()));
+    configurationDto.setWinterTimeDetails(
+        ZonedDateTime.ofInstant(
+            Instant.ofEpochMilli(147L * 24 * 60 * 60 * 1000), ZoneId.systemDefault()));
```
smvdheijden, line 84: Have you considered moving the creation of the dto to a separate method?

jessevdent: Yes I could do that, but since I don't do much more that the create of the dto in the method it doesn't make a big difference. But if you prefer that I can move it ofcourse

osgp/platform/osgp-core/src/test/java/org/opensmartgridplatform/core/infra/jms/protocol/outbound/serialization/ResponseDeserializationTest.java
jessevdent: Same response as in the other project, Yes I could be since we validate it only in one test and don't really reuse it might be a lot of overhead.

